PR #18, opened 2019-05-22
Downgrade typehint for group rules in Validator

KartaviK:
Iterable hint will make it possible to use objects, that implements ArrayAccess interface

| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no
| Fixed issues  |


hiqsol:
Thank you!

PR #21, opened 2019-05-22
 Ip validator

askew-:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | yes
| Tests pass?   | yes?
| Fixed issues  |

### DIFF
--- a/src/Rule/Ip.php
+++ b/src/Rule/Ip.php
@@ -2,7 +2,7 @@
 
 namespace Yiisoft\Validator\Rule;
 
-use Yiisoft\Validator\IpHelper;
+use yii\helpers\IpHelper;
 use Yiisoft\Validator\Result;
 use Yiisoft\Validator\Rule;
 
--- a/src/Rule/Ip.php
+++ b/src/Rule/Ip.php
@@ -2,8 +2,7 @@
 
 namespace Yiisoft\Validator\Rule;
 
-use yii\helpers\IpHelper;
-use Yiisoft\Validator\DataSet;
+use Yiisoft\Validator\IpHelper;
 use Yiisoft\Validator\Result;
 use Yiisoft\Validator\Rule;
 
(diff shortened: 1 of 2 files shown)


src/Rule/Ip.php
```diff
@@ -249,41 +248,82 @@ public function getRanges()
...
+
+        return $this;
+    }
 
+    /**
+     * @param bool $subnet
+     *
+     * @return Ip
+     */
+    public function useSubnet($subnet): self
```
samdark: I'd introduce two methods instead:

- requireSubnet
- optionalSubnet

Also internally it makes sense to introduce three private constants instead of having true/false/null as magic values.

src/Rule/Ip.php
```diff
@@ -249,41 +248,82 @@ public function getRanges()
...
+
+        return $this;
+    }
+
+    /**
+     * @param bool $ipv4
+     *
+     * @return Ip
+     */
+    public function useIpv4($ipv4): self
```
samdark:
```suggestion
    public function allowIpv4(bool $value): self
```

src/Rule/Ip.php
```diff
@@ -249,41 +248,82 @@ public function getRanges()
         return $this->ranges;
     }
 
-    protected function validateValue($value): Result
+    /**
+     * @param bool $ipv6
+     *
+     * @return Ip
+     */
+    public function useIpv6(bool $ipv6): self
```
samdark:
```suggestion
    public function allowIpv6(bool $value): self
```

src/Rule/Ip.php
```diff
@@ -249,41 +248,82 @@ public function getRanges()
...
     {
-        $result = new Result();
+        $this->ipv6 = $ipv6;
```
samdark:
```suggestion
        $this->allowIpv6 = $value;
```

src/Rule/Ip.php
```diff
@@ -249,41 +248,82 @@ public function getRanges()
...
+    }
+
+    /**
+     * @param bool $ipv4
+     *
+     * @return Ip
+     */
+    public function useIpv4($ipv4): self
+    {
+        $this->ipv4 = $ipv4;
```
samdark:
```suggestion
        $this->allowIpv4 = $value;
```

src/Rule/Ip.php
```diff
@@ -249,41 +248,82 @@ public function getRanges()
...
+    }
 
+    /**
+     * @param bool $subnet
+     *
+     * @return Ip
+     */
+    public function useSubnet($subnet): self
+    {
+        $this->subnet = $subnet;
```
samdark:
```suggestion
        $this->subnetStrategy = $subnet;
```

src/Rule/Ip.php
```diff
@@ -249,41 +248,82 @@ public function getRanges()
...
+
+        return $this;
+    }
+
+    /**
+     * @param bool $negation
+     *
+     * @return Ip
+     */
+    public function useNegation(bool $negation): self
```
samdark:
```suggestion
    public function allowNegation(bool $value): self
```

src/Rule/Ip.php
```diff
@@ -249,41 +248,82 @@ public function getRanges()
...
+    {
+        $this->negation = $negation;
```
samdark:
```suggestion
        $this->allowNegation = $value;
```

src/Rule/Ip.php
```diff
@@ -249,41 +248,82 @@ public function getRanges()
...
         }
 
-        $result = $this->validateSubnet($value);
-        if (is_array($result)) {
-            $result[1] = array_merge(['ip' => is_array($value) ? 'array()' : $value], $result[1]);
-            return $result;
-        }
-
-        return $result;
+        return $this->validateSubnet($value);
```
samdark: It makes little sense to have a separate method in this case.

tests/Rule/IpTest.php
```diff
@@ -33,27 +33,27 @@ public function provideRangesForSubstitution()
      */
     public function testRangesSubstitution($range, $expectedRange)
     {
-        $validator = new Ip(['ranges' => $range]);
-        $this->assertEquals($expectedRange, $validator->ranges);
+        $validator = new Ip();
+        $validator->setRanges($range);
```
samdark:
```suggestion
        $validator->ranges($range);
```

kamarton:
Closable via https://github.com/yiisoft/validator/pull/32

PR #22, opened 2019-05-22
Add helpful constructor for Result

KartaviK:
| Q             | A
| ------------- | ---
| Is bugfix?    | no
| New feature?  | yes
| Breaks BC?    | no
| Tests pass?   | yes/no

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -10,6 +10,15 @@ final class Result
 {
     private $valid = true;
     private $errors = [];
+    
+    public function __construct(iterable $messages = [])
+    {
+        if (!empty($messages)) {
+            foreach($messages as $error) {
+                $this->addError($error);
+            }
+        }
+    }
 
     public function isValid(): bool
     {


src/Result.php
```diff
@@ -10,6 +10,15 @@ final class Result
 {
     private $valid = true;
     private $errors = [];
+    
+    public function __construct(iterable $messages = [])
```
samdark: How would you use it?

KartaviK: It can be something like this:

```php
$rules = new Rules([
    function ($value): Result {
        $errors =[] ;

        if ($value !== 41) {
            $errors[] = 'Value should be 41!';
        }
        if ($value !== 42) {
            $errors[] = 'Value should be 42!';
        }
        if ($value !== 43) {
            $errors[] = 'Value should be 43!';
        }

        return new Result($errors);
    }
]);
```

samdark: Not a big gain compared to:

```php
$rules = new Rules([
    function ($value): Result {
       $result = new Result();

        if ($value['first'] === 41) {
            $result->addError('First should not be 41!');
        }
        if ($value['second'] === 42) {
            $result->addError('Second should not be 42!');
        }
        if ($value['third'] === 43) {
            $result->addError('Third should not be 43!');
        }

        return $result;
    }
]);
```

And having two ways of using it makes things a bit more complicated.

PR #25, opened 2019-10-21
Fixed Date Validator & Test for it

evil1:
| Q             | A
| ------------- | ---
| Is bugfix?    | yes
| New feature?  | no
| Breaks BC?    | no
| Tests pass?   | yes

### DIFF
--- a/src/FormatConverterHelper.php
+++ b/src/FormatConverterHelper.php
@@ -7,10 +7,6 @@ use IntlDateFormatter;
  * FormatConverterHelper provides concrete implementation for [[FormatConverter]].
  *
  * Do not use BaseFormatConverter. Use [[FormatConverter]] instead.
- *
- * @author Carsten Brandt <mail@cebe.cc>
- * @author Enrica Ruedin <e.ruedin@guggach.com>
- * @since 2.0
  */
 class FormatConverterHelper
 {
@@ -18,7 +14,7 @@ class FormatConverterHelper
      * @var array the php fallback definition to use for the ICU short patterns `short`, `medium`, `long` and `full`.
      * This is used as fallback when the intl extension is not installed.
      */
-    public static $phpFallbackDatePatterns = [
+    private static $phpFallbackDatePatterns = [
         'short' => [
             'date' => 'n/j/y',
             'time' => 'H:i',
@@ -40,34 +36,8 @@ class FormatConverterHelper
             'datetime' => 'l, F j, Y g:i:sA T',
         ],
     ];
-    /**
-     * @var array the jQuery UI fallback definition to use for the ICU short patterns `short`, `medium`, `long` and `full`.
-     * This is used as fallback when the intl extension is not installed.
-     */
-    public static $juiFallbackDatePatterns = [
-        'short' => [
-            'date' => 'd/m/y',
-            'time' => '',
-            'datetime' => 'd/m/y',
-        ],
-        'medium' => [
-            'date' => 'M d, yy',
-            'time' => '',
-            'datetime' => 'M d, yy',
-        ],
-        'long' => [
-            'date' => 'MM d, yy',
-            'time' => '',
-            'datetime' => 'MM d, yy',
-        ],
-        'full' => [
-            'date' => 'DD, MM d, yy',
-            'time' => '',
-            'datetime' => 'DD, MM d, yy',
-        ],
-    ];
 
-    private static $_icuShortFormats = [
+    private static $icuShortFormats = [
         'short' => 3, // IntlDateFormatter::SHORT,
         'medium' => 2, // IntlDateFormatter::MEDIUM,
         'long' => 1, // IntlDateFormatter::LONG,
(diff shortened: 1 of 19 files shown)


src/Rule/InRange.php
```diff
@@ -105,4 +105,61 @@ public function validateValue($value): Result
...
+     * This method will return `true`, if all elements of `$needles` are contained in
+     * `$haystack`. If at least one element is missing, `false` will be returned.
+     * @param array|\Traversable $needles The values that must **all** be in `$haystack`.
+     * @param array|\Traversable $haystack The set of value to search.
+     * @param bool $strict Whether to enable strict (`===`) comparison.
+     * @throws InvalidArgumentException if `$haystack` or `$needles` is neither traversable nor an array.
+     * @return bool `true` if `$needles` is a subset of `$haystack`, `false` otherwise.
+     * @since 2.0.7
+     */
+    public static function isSubset($needles, $haystack, $strict = false)
```
samdark: That exists in https://github.com/yiisoft/arrays/blob/master/src/ArrayHelper.php#L835

src/Rule/InRange.php
```diff
@@ -105,4 +105,61 @@ public function validateValue($value): Result
...
+     * but additionally works for objects that implement the [[\Traversable]] interface.
+     * @param mixed $needle The value to look for.
+     * @param array|\Traversable $haystack The set of values to search.
+     * @param bool $strict Whether to enable strict (`===`) comparison.
+     * @return bool `true` if `$needle` was found in `$haystack`, `false` otherwise.
+     * @throws InvalidArgumentException if `$haystack` is neither traversable nor an array.
+     * @see https://secure.php.net/manual/en/function.in-array.php
+     * @since 2.0.7
+     */
+    public static function isIn($needle, $haystack, $strict = false)
```
samdark: That exists in https://github.com/yiisoft/arrays/blob/master/src/ArrayHelper.php#L795

src/Rule/Ip.php
```diff
@@ -275,10 +306,15 @@ public function validateValue($value): Result
      * array  - an error occurred during the validation.
      * Array[0] contains the text of an error, array[1] contains values for the placeholders in the error message
      */
-    private function validateSubnet(string $ip)
+    private function validateSubnet(string &$ipAddress)
```
samdark: What is the purpose of & here?

src/Rule/Date.php
```diff
@@ -186,54 +189,83 @@ class Date extends Rule
         'full' => 0, // IntlDateFormatter::FULL,
     ];
 
-    public function __construct()
+    /**
+     * Date constructor.
```
samdark: It was meant to split this class into individual ones. There's `TODO: split into 3 separate classes`

src/Rule/Ip.php
```diff
@@ -25,11 +24,23 @@
  * ['ip_address', 'ip', 'expandIPv6' => true], // expands IPv6 address to a full notation format
  * ```
  *
- * @property array $ranges The IPv4 or IPv6 ranges that are allowed or forbidden. See [[setRanges()]] for
+ * @property array $ranges The IPv4 or IPv6 ranges that are allowed or forbidden. See [[ranges()]] for
  * detailed description.
  */
 class Ip extends Rule
 {
+    const IPV4 = 4;
```
samdark: Visibility for constants should be specified.

src/Rule/Date.php
```diff
@@ -173,7 +174,9 @@ class Date extends Rule
      * If this property is null, the value of [[min]] will be used (before parsing).
      */
     public $minString;
```
samdark: We need to make these non-public. Same for all other properties.

src/Rule/Date.php
```diff
@@ -186,54 +189,83 @@ class Date extends Rule
...
-                $this->format = Yii::getApp()->formatter->datetimeFormat;
-            } elseif ($this->type === self::TYPE_TIME) {
-                $this->format = Yii::getApp()->formatter->timeFormat;
-            } else {
-                throw new InvalidConfigException('Unknown validation type set for DateValidator::$type: ' . $this->type);
+        $this->format = $format;
+        $this->locale = $locale;
+        $this->timeZone = $timeZone;
+
+//        if (!empty($type) && !in_array($type, [self::TYPE_DATE, self::TYPE_DATETIME, self::TYPE_TIME])) {
```
samdark: Should be removed?

src/Rule/Ip.php
evil1: Initially this method returned `array|string`. `Array` in case of error, and `string` with formatted value, in case of success.
Now we return `Result()`. `&` is used to set the formatted value directly to the parameter.

evil1: What shall I do then? Remove unnecessary tests and functionality?

samdark: Yes. See https://forum.yiiframework.com/t/saving-or-killing-non-validation-in-validators/126086

src/FormatConverterHelper.php
```diff
@@ -0,0 +1,303 @@
...
+     * [php date() function format]: https://secure.php.net/manual/en/function.date.php
+     * [ICU format]: http://userguide.icu-project.org/formatparse/datetime#TOC-Date-Time-Format-Syntax
+     *
+     * @param string $pattern date format pattern in ICU format.
+     * @param string $type 'date', 'time', or 'datetime'.
+     * @param string $locale the locale to use for converting ICU short patterns `short`, `medium`, `long` and `full`.
+     * If not given, `Yii::$app->language` will be used.
+     * @return string The converted date format pattern.
+     */
+    public static function convertDateIcuToPhp($pattern, $type = 'date', $locale): string
```
samdark:
```suggestion
    public static function convertDateIcuToPhp(string $pattern, string $type = 'date', string $locale): string
```

src/FormatConverterHelper.php
```diff
@@ -0,0 +1,303 @@
...
+ *
+ * Do not use BaseFormatConverter. Use [[FormatConverter]] instead.
+ */
+class FormatConverterHelper
+{
+    /**
+     * @var array the php fallback definition to use for the ICU short patterns `short`, `medium`, `long` and `full`.
+     * This is used as fallback when the intl extension is not installed.
+     */
+    private static $phpFallbackDatePatterns = [
```
samdark: Could be converted to a constant.

src/FormatConverterHelper.php
```diff
@@ -0,0 +1,303 @@
...
+            'datetime' => 'F j, Y g:i:sA',
+        ],
+        'full' => [
+            'date' => 'l, F j, Y',
+            'time' => 'g:i:sA T',
+            'datetime' => 'l, F j, Y g:i:sA T',
+        ],
+    ];
+
+    private static $icuShortFormats = [
```
samdark: Could be converted to a constant.

src/Rule/Date.php
evil1: Are we using `_` prefix for private properties?

samdark: No.

src/Rule/Ip.php
kamarton: The Ip validator has since been fixed for yii3. PR https://github.com/yiisoft/validator/pull/32

samdark:
Decided to remove Date validator for now. It should be re-considered how it's done overall first.

Thanks for contributing anyway.

PR #32, opened 2019-11-01
Ip validator and Ip tests renovation

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | -
| Alternative PR | https://github.com/yiisoft/validator/pull/21 (abandoned)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,9 +28,6 @@
         "yiisoft/network-utilities": "dev-master",
         "ext-intl": "*"
     },
-    "suggest": {
-        "yiisoft/network-utilities": "For IP and network validations"
-    },
     "require-dev": {
         "phpunit/phpunit": "^7.2"
     },
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,11 @@
         "ext-mbstring": "*",
         "yiisoft/strings": "dev-master",
         "yiisoft/arrays": "dev-master",
+        "yiisoft/network-utilities": "dev-master",
         "ext-intl": "*"
+    },
+    "suggest": {
+        "yiisoft/network-utilities": "For IP and network validations"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.2"
--- a/src/Rule/Ip.php
+++ b/src/Rule/Ip.php
@@ -2,8 +2,7 @@
 
 namespace Yiisoft\Validator\Rule;
 
-use yii\helpers\IpHelper;
-use Yiisoft\Validator\DataSetInterface;
+use Yiisoft\NetworkUtilities\IpHelper;
 use Yiisoft\Validator\Result;
 use Yiisoft\Validator\Rule;
 
@@ -34,7 +33,8 @@ class Ip extends Rule
      * Negation char.
      *
      * Used to negate [[ranges]] or [[networks]] or to negate validating value when [[negation]] is set to `true`.
-     * @see negation
+     *
+     * @see allowNegation
      * @see networks
      * @see ranges
      */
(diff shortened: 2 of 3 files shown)


kamarton:
@samdark questions
- Does it make sense to allow leading zeros for IPv4? (`10.001.000002.1`) https://superuser.com/a/929180/1082613
In my opinion, it makes no sense and only raises additional problems, eg. if allowed, then abnormal IP addresses will also pass through. `0000000000001.0000000000001.000000000001.00000000001`.
- In the original code, the subnet had three states: optional, require, not allow. Since cannot modify the input value, required state does not make sense. `127.0.0.1 === 127.0.0.1 / 32`.

kamarton:
Tests with `FakedValidationModel` style is relevant?

samdark:
> Does it make sense to allow leading zeros for IPv4?

No.

> In the original code, the subnet had three states: optional, require, not allow. Since cannot modify the input value, required state does not make sense. 127.0.0.1 === 127.0.0.1 / 32.

OK.

> Tests with FakedValidationModel style is relevant?

No.

kamarton:
Finally, I left all the subnet options (not allow, reuqired, optional). Reviewable.

Desktop screenshot

kamarton:
The `Ip::getRanges()` function may not be required. It is currently used only for tests, but important for them.

samdark:
Any idea why travis builds fail?

composer.json
```diff
@@ -25,8 +25,12 @@
         "ext-mbstring": "*",
         "yiisoft/strings": "dev-master",
         "yiisoft/arrays": "dev-master",
+        "yiisoft/network-utilities": "dev-master",
         "ext-intl": "*"
     },
+    "suggest": {
```
samdark: If the package is required, there's no sense in suggesting it.

src/Rule/Ip.php
```diff
@@ -222,165 +204,142 @@ public function __construct()
...
      * ```
      *
      * In this example, access is allowed for all the IPv4 and IPv6 addresses excluding the `192.168.10.0/24` subnet.
      * IPv4 address `192.168.10.128` is also allowed, because it is listed before the restriction.
-     * @property array the IPv4 or IPv6 ranges that are allowed or forbidden.
-     * See [[setRanges()]] for detailed description.
+     * @return static
      */
-    public function setRanges($ranges)
+    public function ranges(array $ranges)
```
samdark:
```suggestion
    public function ranges(array $ranges): self
```

src/Rule/Ip.php
```diff
@@ -222,165 +204,142 @@ public function __construct()
...
+    public function getRanges(): array
     {
         return $this->ranges;
     }
 
-    public function validateValue($value): Result
+    /**
+     * @return static
+     */
+    public function allowIpv4(bool $value = true)
```
samdark:
```suggestion
    public function allowIpv4(bool $value = true): self
```

src/Rule/Ip.php
```diff
@@ -222,165 +204,142 @@ public function __construct()
...
 
-        $result = $this->validateSubnet($value);
-        if (is_array($result)) {
-            $result[1] = array_merge(['ip' => is_array($value) ? 'array()' : $value], $result[1]);
-            return $result;
-        }
+    /**
+     * @return static
+     */
+    public function allowIpv6(bool $value = true)
```
samdark:
```suggestion
    public function allowIpv6(bool $value = true): self
```

src/Rule/Ip.php
```diff
@@ -222,165 +204,142 @@ public function __construct()
...
+    {
+        $this->allowIpv6 = $value;
+        return $this;
+    }
 
-        return $result;
+    /**
+     * @return static
+     */
+    public function allowNegation(bool $value = true)
```
samdark:
```suggestion
    public function allowNegation(bool $value = true): self
```

src/Rule/Ip.php
```diff
@@ -222,165 +204,142 @@ public function __construct()
...
-     *
-     * @param $ip string
-     * @return string|array
-     * string - the validation was successful;
-     * array  - an error occurred during the validation.
-     * Array[0] contains the text of an error, array[1] contains values for the placeholders in the error message
+     * @return static
      */
-    private function validateSubnet(string $ip)
+    public function allowSubnet(bool $allow = true, bool $required = false)
```
samdark:
```suggestion
    public function allowSubnet(bool $allow = true, bool $required = false): self
```

samdark: Also it seems to try setting two separate things at the same time. How about `allowSubnet()`, `requireSubnet()`, `disallowSubnet()`?

src/Rule/Ip.php
samdark: How about allowNegation and disallowNegation?

samdark: Same w/ other allow* methods.

kamarton:
> Any idea why travis builds fail?

The other classes in the package are not ready yet, so the tests will fail.

composer.json
kamarton: removed

src/Rule/Ip.php
kamarton: Phpdoc `@return static` is a better solution because `self` must be compliance with the parent class name after inheritance.

example
```php
MyExtendendIp extends Ip {         ˇˇ----- the PHP processor throws an error if I do not use the parent class.
  public function ranges($ranges): Ip {
     // overerride
  }
}
```

Of course, this is not a problem if we set the `Ip` class to `final`.

src/Rule/Ip.php
kamarton: > How about allowNegation and disallowNegation?

like it.

src/Rule/Ip.php
kamarton: > How about allowSubnet(), requireSubnet(), disallowSubnet()?

Like it. It is much clearer to use.

kamarton: added

src/Rule/Ip.php
kamarton: added

kamarton:
depend on https://github.com/yiisoft/network-utilities/pull/13

samdark:
Merged. Thank you!

PR #41, opened 2019-12-25
Fix #28: Implement context aware validation

### LINKED ISSUE #28: Validator requires one attribute and cannot see the whole model
### What steps will reproduce the problem?

In yii2, it was easy to access any of the model's attributes. In yii3, In Yii3 this option seems to me not feasible.

https://www.yiiframework.com/doc/guide/2.0/en/input-validation#standalone-validators

https://github.com/yiisoft/validator/blob/725fdfc3c18d8ef0fee3ed80fd73abe417620d77/src/Validator.php#L36

This issue also affects the `Result` class because it is not suitable for adding errors to an attribute.

### What is the expected result?

I get the whole dataset in an extra parameter, and change return type to `ResultSet`

```php
public function validate($value, DataSetInterface $dataSet): ResultSet
// or yii2 style
public function validate(?string $attribute, DataSetInterface $dataSet): ResultSet
```

https://github.com/yiisoft/validator/blob/725fdfc3c18d8ef0fee3ed80fd73abe417620d77/src/Rule.php#L10

https://github.com/yiisoft/validator/blob/725fdfc3c18d8ef0fee3ed80fd73abe417620d77/src/Result.php#L19

https://github.com/yiisoft/validator/blob/725fdfc3c18d8ef0fee3ed80fd73abe417620d77/src/ResultSet.php#L12

usage:

```php
[
  /* numeric index -> attribute=null */ (new MyComplexRule())->moreParameters()...
]
```

### What do you get instead?

```php
[
  'notRelevantAttributeName' => (new MyComplexRule())->with($this /* or $dataset ??*/)->moreParameters()...
]
```

But in this case, I have no way of returning an arbitrary attribute error of the model.

### Additional info

Also, to keep the benefits of the current solution, the `Rule` class should be somehow allow the` Result` class to be used internally in simpler cases.

### Use case

I need to check for additional attributes depending on a type attribute.

| Q                | A
| ---------------- | ---
| Version          | 3.0.x-dev
| PHP version      | -
| Operating system | -

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ❌
| Fixed issues  | #28


src/Rule.php
```diff
@@ -39,12 +47,13 @@ protected function formatMessage(string $message, array $arguments = []): string
 
     /**
      * @param bool $value
-     * @return $this
+     * @return self
      */
```
roxblnfk: Phpdoc can be removed

src/Rule.php
```diff
@@ -3,20 +3,28 @@
...
 {
-    private $skipOnEmpty = false;
+    private bool $skipOnEmpty = false;
 
-    public function validate($value): Result
+    /**
+     * @param mixed $value
+     * @param DataSetInterface|null $dataSet
+     * @return Result
+     */
```
roxblnfk: More info required or phpdoc can be removed

src/Result.php
```diff
@@ -4,22 +4,25 @@
...
+    private bool $valid = true;
+    private array $errors = [];
 
     public function isValid(): bool
     {
         return $this->valid;
     }
 
-    public function addError(string $message): void
+    public function addError(string $message): self
```
roxblnfk: mb need to rename to `withError`?

src/Rule/Exist.php
```diff
@@ -208,7 +209,7 @@ private function getTargetClass($model)
         return $this->targetClass ?? get_class($model);
     }
 
-    protected function validateValue($value): Result
+    protected function validateValue($value, DataSetInterface $dataSet = null): Result
```
roxblnfk: This method should return `Result` instance.
```php
return $this->valueExists($this->targetClass, $query, $value) ? null : [$this->message, []];
```

src/Rule/File.php
```diff
@@ -204,7 +204,7 @@ private function filterFiles(array $files)
         return $result;
     }
 
-    public function validateValue($value): Result
+    protected function validateValue($value, DataSetInterface $dataSet = null): Result
```
roxblnfk: This method should return `Result` instance.
```php
return [$this->message, []];
```

src/Rule/Each.php
```diff
@@ -161,7 +160,7 @@ public function validateAttribute(DataSetInterface $data, string $attribute): Re
         $model->addErrors([$attribute => $detectedErrors]);
     }
 
-    public function validateValue($value): Result
+    protected function validateValue($value, DataSetInterface $dataSet = null): Result
```
roxblnfk: This method always should return `Result`
On line 182:
```php
return [$this->message, ['value' => $v]];
```

src/Rule/Time.php
```diff
@@ -3,12 +3,13 @@
...
 namespace Yiisoft\Validator\Rule;
 
+use Yiisoft\Validator\DataSetInterface;
 use Yiisoft\Validator\Result;
 use Yiisoft\Validator\Rule;
 
 class Time extends Rule
 {
-    public function validateValue($value): Result
+    protected function validateValue($value, DataSetInterface $dataSet = null): Result
```
roxblnfk: why this method is empty?

src/Rules.php
```diff
@@ -5,49 +5,54 @@
...
+
+        if (!$rule instanceof Rule) {
+            throw new \InvalidArgumentException('Rule should be either instance of Rule class or a callable');
+        }
+
+        return $rule;
     }
 
-    public function getIterator()
+    public function add(Rule $rule): self
```
roxblnfk: mb need rename to `withRule`?

src/ResultSet.php
```diff
@@ -2,38 +2,46 @@
...
 final class ResultSet implements \IteratorAggregate
 {
     /**
      * @var Result[]
      */
-    private $results = [];
+    private array $results = [];
 
-    public function addResult(string $attribute, Result $result): void
+    public function addResult(string $attribute, Result $result): self
```
kamarton: `addResult(Result $result, ?string $attribute = null):self`
- For a primitive type result, the attribute is irrelevant.
- For a dataset type result, the attribute is require

src/Result.php
xepozz: Why it made like a fluent method? Class `Result` is just storage for error messages, not a builder.
I think you need to revert your changes.

samdark: No. It's not setting error. It adds it.

src/Rule.php
```diff
@@ -3,20 +3,28 @@
...
+    final public function validate($value, DataSetInterface $dataSet = null): Result
```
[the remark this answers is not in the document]
samdark: Because Rule validates a single value.

src/Rule.php
```diff
@@ -3,20 +3,28 @@
...
+        if ($this->skipOnEmpty && $this->isEmpty($value)) {
             return new Result();
         }
 
-        return $this->validateValue($value);
+        return $this->validateValue($value, $dataSet);
     }
 
-    abstract public function validateValue($value): Result;
+    abstract protected function validateValue($value, DataSetInterface $dataSet = null): Result;
```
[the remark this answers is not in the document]
samdark: ResultSet is a result of validating whole data set.

src/Result.php
samdark: Yes. You're right.

src/ResultSet.php
samdark: ResultSet is for data set validation results, not primitive results.

samdark:
Thanks for review, everyone!

PR #53, opened 2020-02-25
Translator implementation and refactoring

thenotsoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  |


src/RuleResult.php
```diff
@@ -0,0 +1,29 @@
...
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+/**
+ * Result represents a single value validation result.
+ * It may either be success or contain one or multiple errors.
+ */
+final class RuleResult
```
samdark: It is confusing that there are now both RuleResult and Result...

src/Validator.php
```diff
@@ -1,26 +1,33 @@
...
 
-    /**
-     * Validator constructor.
-     * @param $rules
-     */
-    public function __construct(iterable $rules = [])
-    {
+    public function __construct(
+        iterable $rules = [],
+        TranslatorInterface $translator = null,
```
samdark: I don't think that would work since there is valid usage that doesn't involve validator instance: https://github.com/yiisoft/validator#general-usage

samdark:
As a variant, we can merge RuleResult into Result and require translator when calling `getErrors()`.

src/ResultSet.php
```diff
@@ -13,16 +17,21 @@ final class ResultSet implements \IteratorAggregate
      */
     private array $results = [];
 
-    public function addResult(string $attribute, Result $result): void
-    {
+    public function addResult(
+        string $attribute,
+        Result $result,
+        ?TranslatorInterface $translator = null,
```
samdark: Usually all results within a set have the same translation language, domain and Translator.

samdark:
👍 Thank you!

Issue #29, opened 2019-10-31
yii2 Validator::$skipOnError not implemented


kamarton:
### What steps will reproduce the problem?

https://www.yiiframework.com/doc/api/2.0/yii-validators-validator#$skipOnError-detail

Yii2 gave you the option to skip validations if an error was already registered. This is particularly important in terms of performance.

For example, if check a string for both length and uniqueness in a database, then obviously if the length is not good then the uniqueness test does not make sense.

### What is the expected result?

The rules should be skippable.

### What do you get instead?

All rules are executed.

| Q                | A
| ---------------- | ---
| Version          | 3.0.x-dev
| PHP version      | -
| Operating system | -

samdark:
Could be added to `Rules` and `Validator`.

PR #54, opened 2020-02-27
added "skipOnError" ability & minor changes

thenotsoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | #29


src/Rule.php
```diff
@@ -11,10 +11,15 @@
  */
 abstract class Rule
 {
+    const SKIP_ON_ANY_ERROR = 1;
+    const SKIP_ON_ATTRIBUTE_ERROR = 2;
```
rustamwin:
```suggestion
    public const SKIP_ON_ANY_ERROR = 1;
    public const SKIP_ON_ATTRIBUTE_ERROR = 2;
```

src/ResultSet.php
```diff
@@ -31,6 +35,11 @@ public function addResult(
         }
     }
 
+    public function hasErrors(): bool
```
rustamwin: :+1:

src/Rule.php
thenotsoft: These constants will be removed

src/Rule.php
```diff
@@ -41,22 +47,25 @@ abstract class Rule
      */
     abstract protected function validateValue($value, DataSetInterface $dataSet = null): Result;
 
-    public function setTranslator(TranslatorInterface $translator): self
+    public function withTranslator(TranslatorInterface $translator): self
     {
-        $this->translator = $translator;
-        return $this;
+        $new = clone $this;
```
kamarton: Do you really need to create a new copy? I don't think there is a use-case. If your goal is cascading, you should:
```php
public function translator(TranslatorInterface $translator): self 
{
  $this->translator = $translator;
  return $this;
}

$validator->translator(...)->translatorDomain(...)->translationLocale(...)->...
```

src/Rule.php
```diff
@@ -73,6 +82,13 @@ public function translateMessage(string $message, array $arguments = []): string
         );
     }
 
+    public function skipOnError(bool $value): self
```
kamarton: If you have a new instance: `withSkipOnError()`, but I suggest that you do not have a new instance here:
```php
public function skipOnError(bool $value): self
{
        $this->skipOnError = $value;
        return $this;
}
```

src/Validator.php
```diff
@@ -44,11 +44,10 @@ public function __construct(
     public function validate(DataSetInterface $dataSet): ResultSet
     {
         $results = new ResultSet();
-
         foreach ($this->attributeRules as $attribute => $rules) {
             $results->addResult(
                 $attribute,
-                $rules->validate($dataSet->getAttributeValue($attribute))
+                $rules->validate($dataSet->getAttributeValue($attribute), $dataSet)
```
samdark: This is not related to skip on error but is important bugfix.

src/Rule.php
samdark: @kamarton cloning is cheap. I don't think there's anything bad in making object immutable if it's not meant to have state.

samdark:
👍

PR #58, opened 2020-02-28
Improvements "Callback" rule

thenotsoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |

### DIFF
--- a/src/Rule/Callback.php
+++ b/src/Rule/Callback.php
@@ -16,7 +16,7 @@ class Callback extends Rule
     {
         if (!(is_callable($callback) || (is_array($callback) && count($callback) === 2))) {
             throw new \InvalidArgumentException(
-                'The argument must be a callback or an array with the class and method name.'
+                'The argument must be a callable or an array with the class and method name.'
             );
         }
 
--- a/src/Rule/Callback.php
+++ b/src/Rule/Callback.php
@@ -12,8 +12,14 @@ class Callback extends Rule
 {
     private $callback;
 
-    public function __construct(callable $callback)
+    public function __construct($callback)
     {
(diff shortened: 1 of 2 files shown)


src/Rule/Callback.php
```diff
@@ -12,24 +12,61 @@ class Callback extends Rule
 {
     private $callback;
 
-    public function __construct(callable $callback)
+    public function __construct($callback)
     {
+        if (!(is_callable($callback) || (is_array($callback) && count($callback) === 2))) {
```
xepozz: Why did you do it?
http://sandbox.onlinephpfunctions.com/code/13996f1b47b5103604d7f4263786eb240a31f996

thenotsoft: you're right, `is_callable` will be enough

xepozz: to use private method we can pass the closure like the following:
```php
class Form 
{
    public function validate()
    {
        $rule = new Callback(function ($value) {
            $this->checkValue($value);
        });
        // or shortly
        $rule = new Callback(fn ($value) => $this->checkValue($value));
    }

    private function checkValue(string $value)
    {
        // some check
    }
}
```

samdark:
Thanks!

samdark:
Revered because of https://github.com/yiisoft/validator/pull/58#discussion_r386103133 and potential issues of calling private methods on foreign classes. https://github.com/yiisoft/validator/commit/ddc766d40bd55ca16dbe4620f028daa527588e80

PR #60, opened 2020-02-29
Rule::skipOnEmpty() -> Rule::withSkipOnEmpty()

kamarton:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | -

### DIFF
--- a/src/Rule.php
+++ b/src/Rule.php
@@ -93,7 +93,7 @@ abstract class Rule
      * @param bool $value if validation should be skipped if value validated is empty
      * @return self
      */
-    public function skipOnEmpty(bool $value): self
+    public function withSkipOnEmpty(bool $value): self
     {
         $new = clone $this;
         $new->skipOnEmpty = $value;
--- a/tests/Rule/InRangeTest.php
+++ b/tests/Rule/InRangeTest.php
@@ -31,14 +31,14 @@ class InRangeTest extends TestCase
 
     public function testValidateEmpty(): void
     {
-        $rule = (new InRange(range(10, 20)))->skipOnEmpty(false);
+        $rule = (new InRange(range(10, 20)))->withSkipOnEmpty(false);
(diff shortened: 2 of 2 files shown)


samdark:
I'm not sure about that. Rules are used extensively in the code with many sequential calls:

```php
(new Number())->min(10)->integer()->skipOnEmpty(true);
```

If we're going to rename all the methods to have `with` prefix, that would become:

```php
(new Number())->withMin(10)->withInteger()->withSkipOnEmpty(true);
```

kamarton:
> I'm not sure about that. Rules are used extensively in the code with many sequential calls:
> 
> ```php
> (new Number())->min(10)->integer()->skipOnEmpty(true);
> ```

`min()` and `integer` not clone the object.

Two option:
1. add `with` prefix (this PR)
2. remove object clone

samdark:
> min() and integer not clone the object.

They should.

kamarton:
> They should.

If cloning is needed, then the `with` prefix should be used.

https://github.com/yiisoft/validator/blob/4b54c97e9b0c0c7b68e1c253779ae40c43ace1b0/src/Rule/Number.php#L75-L85

https://github.com/yiisoft/validator/blob/4b54c97e9b0c0c7b68e1c253779ae40c43ace1b0/src/Rule/HasLength.php#L68-L78

...

samdark:
@kamarton `with` prefix will kill readability.

kamarton:
> @kamarton `with` prefix will kill readability.

Currently the method name and hence the usage is inconsistent. `SkipOnEmpty()` returns a new clone (this is already contrary to the `with` prefix used in PSR interfaces). However, the `Number::min()` (and all other Rule methods) method no longer returns a clone.

Whether or not a new clone is needed is a different matter than when the method names do not follow regularity...

samdark:
See https://github.com/yiisoft/validator/pull/61

PR #61, opened 2020-03-02
Make rules immutable

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Tests pass?   | ✔️
| Fixed issues  | #60


kamarton:
It goes against the usual interfaces in PSR, including classes in the yii-web package.

samdark:
That's OK considering that adding `with` is harmful in this case and that method chaining without immutability is a very dangerous practice.

Issue #12, opened 2019-04-24
AtLeastValidator 


samdark:
See https://github.com/yiisoft/yii2/pull/15157

mapeveri:
I sent a PR related to this issue :)

PR #64, opened 2020-04-10
AtLeastValidator

mapeveri:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️

This PR is related to issue #12

### DIFF
--- a/src/Rule/AtLeast.php
+++ b/src/Rule/AtLeast.php
@@ -25,26 +25,14 @@ class AtLeast extends Rule
     private array $attributes = [];
 
     /**
-     * The list of alternative required attributes that will be checked.
-     */
-    private array $alternativeAttributes = [];
-
-    /**
      * Message to display in case of error
      */
     private string $message = 'The model is not valid. Must have at least "{min}" filled attributes.';
 
-    public function __construct(array $data)
+    public function __construct(array $attributes, int $min = 1)
     {
-        foreach ($data as $name => $value) {
-            $this->{$name} = $value;
-        }
-
-        if (empty($this->alternativeAttributes)) {
-            $this->alternativeAttributes = $this->attributes;
-        } else {
-            $this->alternativeAttributes = array_merge($this->attributes, $this->alternativeAttributes);
-        }
+        $this->attributes = $attributes;
+        $this->min = $min;
     }
 
     protected function validateValue($value, DataSetInterface $dataSet = null): Result
@@ -52,7 +40,7 @@ class AtLeast extends Rule
         $valid = false;
         $filledCount = 0;
 
-        foreach ($this->alternativeAttributes as $attribute) {
+        foreach ($this->attributes as $attribute) {
             $filledCount += $this->isEmpty($value->{$attribute}) ? 0 : 1;
         }
 
--- a/src/Rule/AtLeast.php
+++ b/src/Rule/AtLeast.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator\Rule;
+
+use Yiisoft\Validator\Rule;
+use Yiisoft\Validator\Result;
+use Yiisoft\Validator\DataSetInterface;
+
+/**
+ * AtLeastValidator checks if at least $min of many attributes are filled.
+ */
+class AtLeast extends Rule
+{
+    /**
+     * The minimum required quantity of filled attributes to pass the validation.
+     * Defaults to 1.
+     */
+    private int $min = 1;
+
+    /**
+     * The list of required attributes that will be checked.
+     */
+    private array $attributes = [];
+
+    /**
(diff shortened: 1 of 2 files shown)


src/Rule/AtLeast.php
```diff
@@ -0,0 +1,78 @@
...
+    private array $alternativeAttributes = [];
+
+    /**
+     * Message to display in case of error
+     */
+    private string $message = 'The model is not valid. Must have at least "{min}" filled attributes.';
+
+    public function __construct(array $data)
+    {
+        foreach ($data as $name => $value) {
```
xepozz, line 39: I think you should write `$this->attributes = $data` instead

romkatsu: I agree. Allowing the constructor to set any properties(even the parent) via an array is not good.

```php
public function __construct(array $attributes, array $alternativeAttributes = [], int $min = 1)
```

src/Rule/AtLeast.php
```diff
@@ -0,0 +1,78 @@
...
+     * The list of alternative required attributes that will be checked.
+     */
+    private array $alternativeAttributes = [];
+
+    /**
+     * Message to display in case of error
+     */
+    private string $message = 'The model is not valid. Must have at least "{min}" filled attributes.';
+
+    public function __construct(array $data)
```
samdark, line 37: That's not how the rest of validators are configured. Please use the same style.

src/Rule/AtLeast.php
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * The list of required attributes that will be checked.
+     */
+    private array $attributes = [];
+
+    /**
+     * The list of alternative required attributes that will be checked.
+     */
+    private array $alternativeAttributes = [];
```
samdark, line 30: I don't think it's needed. If we need multiple groups we can add multiple validators, can't we?

src/Rule/AtLeast.php
mapeveri: I updated the PR :)

src/Rule/AtLeast.php
mapeveri: I updated the PR, is it ok?.

src/Rule/AtLeast.php
mapeveri: I updated the PR. I have a question related to errrors messages by attribute this line: https://github.com/developeruz/yii2/blob/083da138572d34aae9212d4249d0c64e2cdc2277/framework/validators/AtLeastValidator.php#L102 

I saw that in yii-3 is different. Is necessary this feature?.

src/Rule/AtLeast.php
romkatsu: I'm sorry, I misled you. It is better to do it by analogy with other rules:

```php

    public function attributes(array $attributes): self
    {
        $new = clone $this;
        $new->attributes = $attributes;
        return $new;
    }

    public function min(int $min): self
    {
        $new = clone $this;
        $new->min = $min;
        return $new;
    }

```

And remove the constructor

It is also better to add the message method so that you can change the error text.

src/Rule/AtLeast.php
```diff
@@ -0,0 +1,66 @@
...
+     * The list of required attributes that will be checked.
+     */
+    private array $attributes = [];
+
+    /**
+     * Message to display in case of error
+     */
+    private string $message = 'The model is not valid. Must have at least "{min}" filled attributes.';
+
+    public function __construct(array $attributes, int $min = 1)
```
romkatsu, line 32: It is better to set the min parameter via the method, if necessary.

```php
    public function min(int $value): self
    {
        $new = clone $this;
        $new->min = $value;
        return $new;
    }
```

mapeveri: No problem :)

I updated the PR.

terabytesoftw: > It is better to set the min parameter via the method, if necessary.
> 
> ```
>     public function min(int $value): self
>     {
>         $new = clone $this;
>         $new->min = $value;
>         return $new;
>     }
> ```

Or setter normal why clone ?

romkatsu: > > It is better to set the min parameter via the method, if necessary.
> > ```
> >     public function min(int $value): self
> >     {
> >         $new = clone $this;
> >         $new->min = $value;
> >         return $new;
> >     }
> > ```
> 
> Or setter normal why clone ?

This is how all other validators are made. It is better to stick to one style.

And to call along the chain, as specified in the documentation.

https://github.com/yiisoft/validator/blob/master/README.md

terabytesoftw: you can have setters in stack without needing a copy of the object, for me it doesn't make sense.

samdark:
Thanks!

PR #65, opened 2020-04-12
Add json validator

romkatsu:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  | 

- add json validator
- add trait
- edit composer.json

### DIFF
--- a/src/HasValidationMessage.php
+++ b/src/HasValidationMessage.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+trait HasValidationMessage
+{
+    public function message(string $message): self
+    {
+        $new = clone $this;
+        $new->message = $message;
+        return $new;
+    }
+}
--- a/src/Rule/Json.php
+++ b/src/Rule/Json.php
@@ -17,7 +17,7 @@ class Json extends Rule
 {
     use HasValidationMessage;
 
-    private string $message = 'The value is not json';
+    private string $message = 'The value is not JSON.';
 
     protected function validateValue($value, DataSetInterface $dataSet = null): Result
     {
--- a/src/Rule/Json.php
+++ b/src/Rule/Json.php
@@ -8,7 +8,7 @@ use Yiisoft\Validator\HasValidationMessage;
 use Yiisoft\Validator\Rule;
 use Yiisoft\Validator\Result;
 use Yiisoft\Validator\DataSetInterface;
-use Exception;
+use JsonException;
 
 /**
  * JsonValidator validates that the attribute value is a valid json
@@ -38,9 +38,10 @@ class Json extends Rule
 
         try {
             json_decode($value, false, 512, JSON_THROW_ON_ERROR);
-            return true;
-        } catch (Exception $e) {
+        } catch (JsonException $e) {
             return false;
         }
+
+        return true;
     }
 }
--- a/src/Rule/Json.php
+++ b/src/Rule/Json.php
@@ -17,7 +17,7 @@ class Json extends Rule
 {
     use HasValidationMessage;
 
-    protected string $message = 'The value is not json';
+    private string $message = 'The value is not json';
 
     protected function validateValue($value, DataSetInterface $dataSet = null): Result
     {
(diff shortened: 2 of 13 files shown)


src/HasValidationMessage.php
```diff
@@ -0,0 +1,15 @@
+<?php
```
xepozz, line 1: I don't sure that this trait is really useful. Maybe we can place the `message` method into parent `Rule` instead?

src/Rule/Json.php
```diff
@@ -0,0 +1,46 @@
...
+    private function isValidJson($value): bool
+    {
+        if (!is_string($value)) {
+            return false;
+        }
+
+        try {
+            json_decode($value, false, 512, JSON_THROW_ON_ERROR);
+            return true;
+        } catch (Exception $e) {
```
xepozz, line 42:
```suggestion
        } catch (\JsonException $e) {
```

src/HasValidationMessage.php
romkatsu: Yes, we can. But not all validators will use it. So I put it in the trait.

src/Rule/Json.php
```diff
@@ -0,0 +1,46 @@
...
+
+    private function isValidJson($value): bool
+    {
+        if (!is_string($value)) {
+            return false;
+        }
+
+        try {
+            json_decode($value, false, 512, JSON_THROW_ON_ERROR);
+            return true;
```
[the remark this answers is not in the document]
romkatsu: done

tests/Rule/JsonTest.php
```diff
@@ -0,0 +1,51 @@
...
+        $this->assertFalse($val->validate(10)->isValid());
+        $this->assertFalse($val->validate(null)->isValid());
+    }
+
+    public function testValidValueValidate(): void
+    {
+        $this->assertTrue((new Json())->validate('{"name": "tester"}')->isValid());
+    }
+
+    public function testValidationMessage(): void
```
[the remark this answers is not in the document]
romkatsu: done

src/HasValidationMessage.php
xepozz: Which ones?

src/Rule/Json.php
romkatsu: done

src/HasValidationMessage.php
romkatsu: Callback, CompareTo, Ip, Number

xepozz: Ok thanks

romkatsu:
[Json validator](https://trello.com/c/hJAlXABg/58-json-validator)

samdark:
👍

armpogart:
Not sure if it really matters but I have a benchmark (if I implemented it correctly) that shows that using regex for validating json is faster, and the bigger the json, the more is the difference. For real world case the difference would not be visible at all, but another disadvantage of using `json_decode` for validation purposes here is that it will also consume memory by creating all that objects for the decoded data.

See [benchmark](http://sandbox.onlinephpfunctions.com/code/afc77f800197659391c3d1ea2a02b5e1884b9a4f)

armpogart:
Also `json_decode` successfully decodes any number (which you have check for), simple strings and boolean true (which also don't pass `is_string` check), which is not valid json in most languages. But simple strings still pass you validation, but they are not valid for javascript `JSON.parse`.

That's due to the fact that PHP implements a superset of JSON specified in [RFC 7159](https://www.ietf.org/rfc/rfc7159.txt) vs older standard that is implemented by javascript and majority: [RFC 4267](https://www.ietf.org/rfc/rfc4627.txt).

Also there is a newer revision of json standard: [RFC 8259](https://tools.ietf.org/html/rfc8259), which I doubt any language implemented at all.

samdark:
@armpogart are you sure it is complete regex that validates JSON?

armpogart:
@samdark I'll open another discussion (issue) with details posted here and references to all the data I got. There are also json validation test cases on [json.org](https://www.json.org/json-en.html) which can be used for json validator testing.

Anyways I'll post all the details in an issue reference it here. I think there are some points that need further discussion.

samdark:
Alright. Overall benchmarks look good enough to try it.

PR #66, opened 2020-04-12
Update Rule.php

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Tests pass?   | ✔️

### DIFF
--- a/src/Rule.php
+++ b/src/Rule.php
@@ -76,15 +76,15 @@ abstract class Rule
         return $new;
     }
 
-    public function translateMessage(string $message, array $arguments = []): string
+    protected function translateMessage(string $message, array $params = []): string
     {
         if ($this->translator === null) {
-            return $this->formatMessage($message, $arguments);
+            return $this->formatMessage($message, $params);
         }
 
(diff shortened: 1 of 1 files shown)


src/Rule.php
```diff
@@ -76,15 +76,15 @@ public function translationLocale(string $locale): self
         return $new;
     }
 
-    public function translateMessage(string $message, array $arguments = []): string
+    protected function translateMessage(string $message, array $params = []): string
```
samdark, line 79: I think it should be arguments in this case. See https://trello.com/c/R6DBDMdj/59-make-it-consistent-about-naming-parameters-vs-arguments

xepozz: It isn't just parameters or arguments of function, it parameters of message.
Also the second argument in `Translator::translate` named as `parameters` https://github.com/yiisoft/i18n-translator/blob/master/src/Translator.php#L59

samdark:
👍

Issue #70, opened 2020-06-29
Add the possibility of addRules() in Validator, outside the constructor.


terabytesoftw:
| Q                | A
| ---------------- | ---
| Version          | 3.0
| PHP version      | 7.4.7
| Operating system | Windows 10

PR #72, opened 2020-06-30
Enhacement

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | https://github.com/yiisoft/validator/issues/70

- [x] Add the possibility of addRules() in Validator, outside the constructor. 
- [x] Swicht github-actions.


src/Validator.php
```diff
@@ -30,15 +30,6 @@ public function __construct(
         $this->translator = $translator;
         $this->translationDomain = $translationDomain;
         $this->translationLocale = $translationLocale;
-
-        foreach ($rules as $attribute => $ruleSets) {
```
viktorprogger, line 34: You should delete the `rules` parameter too

terabytesoftw: done, thks.

thenotsoft:
I think that you need to return back the ability to add rules in the constructor, since in current implementations, getting a validator through DI is impossible.

PR #73, opened 2020-07-09
Add "ValidatorFactory" and cleanup

thenotsoft:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Tests pass?   | ✔️
| Fixed issues  |

### DIFF
--- a/src/Rules.php
+++ b/src/Rules.php
@@ -5,34 +5,45 @@ declare(strict_types=1);
 namespace Yiisoft\Validator;
 
 use Yiisoft\Validator\Rule\Callback;
-use Yiisoft\I18n\TranslatorInterface;
 
 /**
  * Rules represents multiple rules for a single value
  */
-class Rules
+final class Rules
 {
     /**
      * @var Rule[]
      */
     private array $rules = [];
-    private ?TranslatorInterface $translator;
-    private ?string $translationDomain;
-    private ?string $translationLocale;
 
-    public function __construct(
-        iterable $rules = [],
-        ?TranslatorInterface $translator = null,
-        ?string $translationDomain = null,
-        ?string $translationLocale = null
-    ) {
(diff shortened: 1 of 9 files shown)


src/Validation.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+use Yiisoft\I18n\TranslatorInterface;
+use Yiisoft\Validator\Rule\Callback;
+
+final class Validation implements ValidationInterface
```
romkatsu, line 10: It looks like a factory. Can be named ValidatorFactory?

terabytesoftw:
What do you think instead of putting the TranslatorInterface in the null constructor, create a setter for your configuration.

thenotsoft:
> What do you think instead of defined the TranslatorInterface in the null constructor, create a setter for your configuration.
> 
> ```
> public function translator(TranslatorInterface $translatorInterface, string $category)
> {
> }
> ```
> 
> Locale is already defined in the interface.

I think for factory it’s more convenient in the constructor, locale is defined in Translator, but if you need translate other locale dynamic ?

src/Rules.php
```diff
@@ -5,72 +5,35 @@
 namespace Yiisoft\Validator;
 
 use Yiisoft\Validator\Rule\Callback;
-use Yiisoft\I18n\TranslatorInterface;
 
 /**
  * Rules represents multiple rules for a single value
  */
-class Rules
+final class Rules
```
xepozz, line 12: Why?

samdark: Do you have a use-case for extending it?

samdark:
👍

PR #76, opened 2020-08-10
Use iterable instead of two subtypes.

armpogart:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

Use `iterable` type (introduced in PHP 7.1) instead of two subtypes (`array` and `\Traversable`). This simplifies the code and allows correct type-hinting without init validation.


samdark:
👍

Issue #43, opened 2019-12-26
Consider clientside validation


samdark:
Need to think about possible ways to implement clientside validation.

1. It could be done in pure JS or any JS framework.
2. This library should not have any JS.

Likely that could be introducing a way to convert rules to array that later could be JSON-ized and passed to client side.

mamadza:
I think one of most interesting feature in yii2 is clientside validation , It's more interesting to have this in yii3 to any way , with more options

xepozz:
If it will be implemented like a fully separeted package then ok.
It would be cool to have 3 packages of validators: 
1. yiisoft/validator
  Fully native-php validator
2. yiisoft/js-validator
  Fully native-js validator
3. yiisoft/js-validator-bridge
  Bridge between previous packages.

But I bet that JS world have own package for validators.
We can just implement bridge to it. It would be greatful solution for now.

kamarton:
One of the main goals could be for the javascript validator to useable with multiple javascript frameworks. I'm thinking of solving it so that it can be solved with vue.js, jquery and future libs. 

It might be a solution to include a validation descriptor in the HTML tag, e.g. `<form data-rules="{descriptor}">` or `<input data-rule="{descriptor}">`.

It might be advisable to treat javascript pages completely separately, I also think about error messages. Javascript is hackable, so for final validation it is trusted by the PHP side, which can only be accepted by form-uploads after ajax or POST.

samdark:
It is not the job of the validator package to generate HTML so instead it can return something like

```php
[
    'attribute1' => [
        'number',
        'max' => 10,
    ],
    'attribute2' => [
        'required'
    ],
]
```

kamarton:
> It is not the job of the validator package to generate HTML so instead it can return something like

I did not intend to generate HTML tags. Basically I think it is necessary to decide where the descriptor is loaded from, since it can affect its processing and thus the code it generates.

kamarton:
The client-side rule is optional, so I suggests defining an interface for `Rule` and `Validator` classes.

samdark:
Would using `asOptions(): array` do?

kamarton:
> Would using `asOptions(): array` do?

What happens if Rule does not support client-side validation? e.g. check domain by reverse DNS. In this case, return with an empty array?

kamarton:
how about `asDescriptor()`? The return value could include type identification as well as options. e.g.
```php
[
[...]

PR #81, opened 2020-09-03
Creating array of data for usage on frontend (client side validation)

darkdef:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️/❌
| Fixed issues  | #43


src/Rule/Email.php
```diff
@@ -126,4 +126,28 @@ public function enableIDN(bool $enableIDN): self
...
+    }
+
+    /**
+     * @inheritDoc
+     * @return array
+     */
+    public function getOptions(): array
+    {
+        return array_merge(
+            $this->allowName ? ['allowName' => true] : [],
```
thenotsoft, line 146: Why if the bool value does not always output it, but return an empty array?

eg `['checkDNS' => $this->checkDNS]`

darkdef: If the default value for a flag has not been changed and this flag has not been changed, the result is empty.
If it is formed for the default values, the returned array will be too large.
We do not force the user to specify all values when creating a validator.
;-)

src/Rule/Url.php
```diff
@@ -112,6 +121,23 @@ public function schemes(array $schemes): self
...
+    {
+        return 'url';
+    }
+
+    public function getOptions(): array
+    {
+        return array_merge(
+            ['message' => $this->translateMessage($this->message)],
+            $this->enableIDN ? ['enableIDN' => true] : [],
+            $this->schemesIsChanged ? ['validSchemes' => $this->validSchemes] : [],
```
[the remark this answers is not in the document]
darkdef: ok

src/Rule/Email.php
darkdef: It's be corrected

darkdef:
How about this resolve:
In class Rule adding method getDefaultOptions for return values by default and add abstract method getOptions (for required to implementation)
```php
    /**
     * Returns rule options.
     * @return array
     */
    abstract public function getOptions(): array ;

    /**
     * Returns default rule options.
     * @return array
     */
    protected function getDefaultOptions(): array
    {
        return [
            'skipOnEmpty' => $this->skipOnEmpty,
            'skipOnError' => $this->skipOnError,
        ];
    }
```
In child classes 
```php
    {
        return array_merge(
            ['min' => $this->min],
            ['message' => $this->translateMessage($this->message, ['min' => $this->min])],
            parent::getDefaultOptions()
        );
    }
```

thenotsoft:
I would prefer this option

```
abstract class Rule implements \JsonSerializable {
    ...

    public function jsonSerialize(): array
    {
        return array_merge(
            $this->getDefaultOptions(), 
            $this->getOptions()
        );
    }

    /**
     * Returns rule options.
     * @return array
     */
    abstract public function getOptions(): array ;

    /**
     * Returns default rule options.
     * @return array
     */
    protected function getDefaultOptions(): array
    {
        return [
            'skipOnEmpty' => $this->skipOnEmpty,
            'skipOnError' => $this->skipOnError,
        ];
    }
    ...
}
```

In child class

```
    {
        return [
             'min' => $this->min,
             'message' => $this->translateMessage($this->message, ['min' => $this->min])
        ];
    }
```

darkdef:
Well, if be use api with XML?
Need solution is required with the correct result (as an array) from a single function.

thenotsoft:
> Well, if be use api with XML?
> Need solution is required with the correct result (as an array) from a single function.
> And it violates SRP

How does this option interfere with xml? there is no difference, calling `getOptions` or `jsonSerialize` both return an array, only in the second case you can immediately use json_encode without additional effort.

What is the SRP violation?

darkdef:
Imho class Validate must not be linked with JsonSerializable

samdark:
I agree with @darkdef. Validator should not be responsible for serialization of options.

xepozz:
Shouldn't  `getName` and `getOptions` be provided with usage an interface instead of abstract Rule?

darkdef:
> 
> 
> Shouldn't `getName` and `getOptions` be provided with usage an interface instead of abstract Rule?

We need to return private properties from the parent class, so the getOptions implementation is present in the parent class.
When using the interface in such a situation, the child classes will still not need to implement the getOptions method.
--- 
Нам необходимо вернуть приватные свойства из родительского класса, поэтому реализация getOptions присутствует в родительском классе.
При использовании интерфейса в такой ситуации, от дочерних классов, всё равно не потребуется реализация метода getOptions

darkdef:
In last commit:
Swaped parent::getOptions with child options.
Added getOptions to GroupRule.
Added test to Validator on create with incorrect rule

xepozz:
Can you please add a basic definition of the name, for example `static::class`?

darkdef:
> Can you please add a basic definition of the name, for example `static::class`?
And extract className from fully qualified name with namespace?
With usage explode/pcre/reflection ?

See my commits - this was the first implementation ;-)
```php
    public function getName(): string
    {
        return strtolower((new \ReflectionClass($this))->getShortName());
    }
```
But @samdark against. Keep it simple, no magic

xepozz:
Why I should to implement `getName` method if my validator will never be used in the frontend?

xepozz:
> And extract className from fully qualified name with namespace?
With usage explode/pcre/reflection ?

No, just fully qualified name.

src/Rule/Number.php
```diff
@@ -114,4 +114,34 @@ private function getNotANumberMessage(): string
         }
         return 'Value must be a number.';
     }
+
+    /**
+     * @param mixed $value the data value to be checked.
+     */
+    private function isNotNumber($value): bool
```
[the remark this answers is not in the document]
darkdef: Yes, this unused. Cutted

samdark (approved):

samdark:
Thanks!

PR #86, opened 2020-10-05
Rename Rule to AbstractRule

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  |


samdark:
I don't think this renaming adds to clarity. Let's keep shorter name.

PR #87, opened 2020-10-07
Refactor concept

komandakycto:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | -

This concept implements the composite pattern and was made as part of the [stream](https://www.youtube.com/watch?v=ucO7229o8Ew&ab_channel=SkyengITeam) . The main focus of the refactoring was to isolate an interface for a rule class and implement that interface in a rule and rule group.

Benefits:
- Work with Rules as is a single Rule
- Set skipOnError and skipOnEmpty options for group of rules
- Reduced cyclomatic complexity of Rule class
- Translation logic moved to Error object
- when() moved to own Rule
- All tests pass


src/Error.php
```diff
@@ -0,0 +1,39 @@
...
+
+    public function getErrors(): array
+    {
+        return array_map(function ($error) {
+            [$message, $parameters] = $error;
+            return $this->translateMessage($message, $parameters);
+        }, $this->errors);
+    }
+
+    public function getRawErrors()
```
andrew-demb, line 35: I think this method should have an `array` return type and described about array items structure in phpdoc, because it is highly used in this way:
```php
foreach ($error->getRawErrors() as $errorItem) {
    [$message, $parameters] = $errorItem;
    // ...
}
```

samdark:
@komandakycto thanks again for ideas. Many were applied to current code with adjustments. Closing the PR since it's not mergeable as is.

PR #91, opened 2020-11-12
Fix phpdocs

Fantom409:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Rule/Ip.php
+++ b/src/Rule/Ip.php
@@ -282,6 +282,7 @@ class Ip extends Rule
      *
      * In this example, access is allowed for all the IPv4 and IPv6 addresses excluding the `192.168.10.0/24` subnet.
      * IPv4 address `192.168.10.128` is also allowed, because it is listed before the restriction.
+     * @return self
      */
     public function ranges(array $ranges): self
     {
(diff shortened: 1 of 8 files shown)


src/Rule/Ip.php
```diff
@@ -382,6 +383,8 @@ public function disallowSubnet(): self
     /**
      * The method checks whether the IP address with specified CIDR is allowed according to the [[ranges]] list.
      *
+     * @param string $ip
```
samdark, line 386: hat doesn't add anything.

src/Rule/Ip.php
```diff
@@ -282,6 +282,7 @@ protected function validateValue($value, DataSetInterface $dataSet = null): Resu
      *
      * In this example, access is allowed for all the IPv4 and IPv6 addresses excluding the `192.168.10.0/24` subnet.
      * IPv4 address `192.168.10.128` is also allowed, because it is listed before the restriction.
+     * @return self
```
samdark, line 285: hat doesn't add anything.

src/ValidatorFactory.php
```diff
@@ -40,7 +40,8 @@ private function normalizeRules(array $rules)
     }
 
     /**
-     * @param Rule|callable
+     * @param Rule|callable $rule
+     * @return Rule
```
samdark, line 44: hat doesn't add anything.

samdark (changes requested):

PR #94, opened 2020-11-22
EmailValidator with IDN fails on hyphen

mapeveri:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #2


samdark:
It seems to break tests: https://github.com/yiisoft/validator/runs/1438130363

src/Rule/Email.php
```diff
@@ -93,6 +94,10 @@ protected function validateValue($value, DataSetInterface $dataSet = null): Resu
             }
         }
 
+        if ($this->enableIDN && $valid === false) {
+            $valid = filter_var($originalValue, FILTER_VALIDATE_EMAIL) ? true : false;
```
samdark, line 98:
```suggestion
            $valid = (bool) filter_var($originalValue, FILTER_VALIDATE_EMAIL);
```

mapeveri:
@samdark Yep, the problem is with this emails:

image

I commeted the email with problems in the test. Any idea why this email passed the filter_var?

Thanks

src/Rule/Email.php
mapeveri: I updated this fix

samdark:
@mapeveri seems that filter_var isn't suitable for the purpose: https://3v4l.org/4oslP

samdark:
filter_var is RFC 5321 and we need way stricter validation.

mapeveri:
@samdark What validation do you recommend?

samdark:
Regular expressions.

mapeveri:
@samdark thanks. I updated this PR :)

src/Rule/Email.php
```diff
@@ -93,6 +94,11 @@ protected function validateValue($value, DataSetInterface $dataSet = null): Resu
             }
         }
 
+        if ($this->enableIDN && $valid === false) {
+            $pattern = '/^([a-zA-Z0-9._%+-]+)@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.)|(([\w-]+\.)+))([a-zA-Z]{2,4}|[0-9]{1,3})(\]?)$/';
```
samdark, line 98: That could be moved to a customizable property same as other regex patterns.

mapeveri: @samdark there is a private pattern regex, maybe a new property $patternIdnEmail as a public property, is ok?

mapeveri: @samdark i upated the PR ;)

mapeveri: @samdark is it ok?

samdark (approved):

samdark:
Thank you!

src/Rule/Email.php
samdark: Yes.

PR #96, opened 2021-01-08
Extract interfaces from Rule

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |


vjik:
Maybe name interfaces without `Validator`?
`ParametrizedRuleInterface`, `TranslatableRuleInterface` and etc.

src/FormatableRuleInterface.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+/**
+ * Rule represents a single value validation rule.
```
samdark, line 8:
```suggestion
 * Formattable rule constructs error messages from a pattern and parameters. The interface allows to set formatter to use.
```

src/ParametrizedRuleInterface.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+/**
+ * Rule represents a single value validation rule.
+ */
+interface ParametrizedRuleInterface extends RuleInterface
```
samdark, line 10:
```suggestion
/**
 * The interface should be implemented for a rule that is meant to be converted to array of name => parameters.
 * Such array is usually passed to the client and then used for client-side validation.
 */
interface ParametrizedRuleInterface extends RuleInterface
```

samdark (approved):

samdark:
👍

PR #97, opened 2021-01-09
Make concept of stateless validator

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️

I want to make `Validator` stateless.
It will avoid to create instance every time when we want to validate another dataset or rules.
Also it will be a little bit user-friendly when we want to get instance of `Validator` with usage DI:
```php
public function action(ValidatorInterface $validator) 
{
    $result = $validator->validate($request, $rules);
}
```
vs
```php
public function action(ValidatorFactoryInterface $validatorFactory) 
{
    $validator = $validatorFactory->create($rules);
    $result = $validator->validate($request);
}
```


src/ValidatorFactory.php
```diff
@@ -5,7 +5,6 @@
 namespace Yiisoft\Validator;
 
 use Yiisoft\Translator\TranslatorInterface;
-use Yiisoft\Validator\Rule\Callback;
 
 final class ValidatorFactory implements ValidatorFactoryInterface
```
samdark, line 9: Wasn't the plan to remove factories?

tests/ValidatorTest.php
```diff
@@ -63,179 +49,27 @@ public function testAddingRulesViaConstructor(): void
...
+        ]);
 
         $this->assertTrue($results->getResult('bool')->isValid());
 
         $intResult = $results->getResult('int');
         $this->assertFalse($intResult->isValid());
         $this->assertCount(1, $intResult->getErrors());
     }
-
-    public function testAsArray(): void
```
samdark, line 115: AsArray should be present either as part of validator or as separate class.

samdark:
Overall 👍

samdark:
Needs to be adjsuted for https://github.com/yiisoft/validator/pull/106

samdark:
👍

PR #99, opened 2021-01-10
Other usage of translator in validator (concept, draft)

darkdef:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️


src/Result.php
```diff
@@ -13,13 +15,18 @@ public function isValid(): bool
         return $this->errors === [];
     }
 
-    public function addError(string $message): void
+    public function addError(string $message, ?array $params = null): void
```
samdark, line 18:
```suggestion
    public function addError(string $message, array $params = []): void
```

src/ErrorMessage.php
```diff
@@ -0,0 +1,64 @@
...
+use Yiisoft\Translator\TranslatorInterface;
+
+final class ErrorMessage
+{
+    public string $message = '';
+    public array $params = [];
+
+    private ?TranslatorInterface $translator;
+
+    public function __construct(string $message, ?array $params = null, ?TranslatorInterface $translator = null)
```
samdark, line 16:
```suggestion
    public function __construct(string $message, array $params = [], ?TranslatorInterface $translator = null)
```

src/ErrorMessage.php
```diff
@@ -0,0 +1,64 @@
...
+    {
+        $this->message = $message;
+        $this->params = $params?? [];
```
samdark, line 19:
```suggestion
        $this->params = $params;
```

src/ErrorMessage.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->translator = $translator;
+    }
+
+    public function withTranslator(?TranslatorInterface $translator = null): self
+    {
+        if ($translator !== null) {
```
samdark, line 25: There's no need for this condition. It prevents removing of translator.

src/Rule/Each.php
```diff
@@ -20,7 +21,7 @@ class Each extends Rule
     private Rules $rules;
 
     private string $incorrectInputMessage = 'Value should be array or iterable';
-    private string $message = '{error} {value} given.';
+    private string $message = ' {value} given.';
```
samdark, line 24: Error should be parameter.

samdark:
Overall good concept 👍

src/ErrorMessage.php
darkdef: I don't understand. If remove this line - $this->translator not be used with empty param $translator

samdark: 1. The method accepts `null`. I'd suppose it means "use no translator".
2. If I'll do `(new ErrorMessage('bla', [], $translator))->withTranslator(null)` with current code it won't remove translator. It will stil return same instance with `$translator` assigned.

src/ValidatorFactory.php
```diff
@@ -10,17 +10,11 @@
...
 {
     private ?TranslatorInterface $translator;
-    private ?string $translationDomain;
-    private ?string $translationLocale;
 
     public function __construct(
-        TranslatorInterface $translator = null,
-        string $translationDomain = null,
-        string $translationLocale = null
+        TranslatorInterface $translator = null
```
xepozz, line 15: Looks like not used dependency

darkdef: Yep, removed

src/Rule/Email.php
```diff
@@ -59,6 +60,8 @@ protected function validateValue($value, DataSetInterface $dataSet = null): Resu
     {
         $originalValue = $value;
         $result = new Result();
+        /** @var ?array $matches */
+        $matches = null;
```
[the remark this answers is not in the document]
darkdef: It's only for psalm analyse. Plesae read the doc for function preg_match.

xepozz (approved):

tests/FormatterMock.php
```diff
@@ -0,0 +1,35 @@
+<?php
```
xepozz, line 1: There is no need to create abstract test case class. Make separated class-factory for this purpose.

src/Result.php
```diff
@@ -6,20 +6,34 @@
...
+    /**
+     * @return ErrorMessage[]
+     */
+    public function getErrors(?ErrorMessageFormatterInterface $formatter = null): array
     {
+        if ($formatter instanceof ErrorMessageFormatterInterface) {
+            return array_map(
+                function ($error) use ($formatter) {
+                    return new ErrorMessage($error->getMessage(), $error->getParameters(), $formatter);
+                },
```
xepozz, line 33:
```suggestion
                $error-> getFormattedMessage($formatter)
```

src/Result.php
```diff
@@ -6,20 +6,34 @@
...
-    public function addError(string $message): void
+    public function addError(ErrorMessage $error): void
     {
-        $this->errors[] = $message;
+        $this->errors[] = $error;
     }
 
-    public function getErrors(): array
+    /**
+     * @return ErrorMessage[]
```
xepozz, line 25:
```suggestion
     * @return string[]
```

src/ErrorMessage.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Yiisoft\Validator;
+
+final class ErrorMessage
+{
+    private string $message = '';
+    private array $parameters = [];
+    private ErrorMessageFormatterInterface $formatter;
+
+    public function __construct(string $message, array $parameters = [], ?ErrorMessageFormatterInterface $formatter = null)
```
xepozz, line 13:
```suggestion
    public function __construct(string $message, array $parameters = [])
```

src/ErrorMessage.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        $this->message = $message;
+        $this->parameters = $parameters;
+        $this->formatter = $formatter ?? new ErrorMessageFormatter();
```
xepozz, line 17:
```suggestion
        $this->formatter = new ErrorMessageFormatter();
```

src/ErrorMessage.php
xepozz: Let's leave formatter only in `getFormattedMessage` method.

src/Rule/Boolean.php
```diff
@@ -78,20 +80,21 @@ protected function validateValue($value, DataSetInterface $dataSet = null): Resu
         return $result;
     }
 
-    public function getOptions(): array
+    public function getOptions(?ErrorMessageFormatterInterface $formatter = null): array
```
xepozz, line 83: I suggest to you to remove that parameter and find and translate ErrorMessage outside of that method.

xepozz (commented):
This PR doesn't have any points without working code or concept of bridge for translating these messages.
Can you make PR into, for example, yii-demo?

src/Result.php
darkdef: No, this method return array with instances of ErrorMessage

tests/FormatterMock.php
xepozz: Why it was resolved without any comments?

src/Result.php
xepozz: The same

darkdef: It's function getRawErrors()

tests/FormatterMock.php
darkdef: resolved

PR #104, opened 2021-01-19
Add methods to ResultSet

romkatsu:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | https://github.com/yiisoft/request-model/issues/16

### DIFF
--- a/src/ResultSet.php
+++ b/src/ResultSet.php
@@ -54,16 +54,16 @@ final class ResultSet implements IteratorAggregate
         return new ArrayIterator($this->results);
     }
 
-    public function getErrors(): self
+    public function getErrors(): array
     {
-        $resultSet = new self();
+        $errors = [];
         foreach ($this->results as $attribute => $result) {
             if (!$result->isValid()) {
-                $resultSet->addResult($attribute, $result);
+                $errors[$attribute] = $result;
             }
         }
 
-        return $resultSet;
+        return $errors;
     }
 
     public function isValid(): bool
--- a/src/ResultSet.php
+++ b/src/ResultSet.php
@@ -68,6 +68,6 @@ final class ResultSet implements IteratorAggregate
 
     public function hasErrors(): bool
     {
-        return $this->hasErrors === false;
+        return $this->hasErrors;
     }
 }
--- a/src/ResultSet.php
+++ b/src/ResultSet.php
@@ -48,4 +48,27 @@ final class ResultSet implements IteratorAggregate
     {
         return new ArrayIterator($this->results);
     }
+
+    public function getErrors(): self
+    {
+        $resultSet = new self();
+        foreach ($this->results as $attribute => $result) {
+            if (!$result->isValid()) {
+                $resultSet->addResult($attribute, $result);
+            }
+        }
+
+        return $resultSet;
+    }
+
+    public function isValid(): bool
+    {
+        foreach ($this->results as $result) {
+            if (!$result->isValid()) {
+                return false;
+            }
+        }
+
+        return true;
+    }
 }
--- a/src/ResultSet.php
+++ b/src/ResultSet.php
@@ -18,18 +18,23 @@ final class ResultSet implements IteratorAggregate
      * @var Result[]
      */
     private array $results = [];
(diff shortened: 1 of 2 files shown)


src/ResultSet.php
```diff
@@ -48,4 +48,27 @@ public function getIterator(): ArrayIterator
...
+            }
+        }
+
+        return $resultSet;
+    }
+
+    public function isValid(): bool
+    {
+        foreach ($this->results as $result) {
+            if (!$result->isValid()) {
```
xepozz, line 67: Please use `addResult` method to catching an added new errored result:
If an added result has errors then whole ResultSet will have an error.

src/ResultSet.php
```diff
@@ -48,4 +48,27 @@ public function getIterator(): ArrayIterator
...
+                return false;
+            }
+        }
+
+        return true;
```
xepozz, line 72:
```suggestion
        return $this->isValid;
```

src/ResultSet.php
```diff
@@ -48,4 +48,27 @@ public function getIterator(): ArrayIterator
...
     }
+
+    public function getErrors(): self
+    {
+        $resultSet = new self();
+        foreach ($this->results as $attribute => $result) {
+            if (!$result->isValid()) {
+                $resultSet->addResult($attribute, $result);
+            }
+        }
```
xepozz, line 59: Are you sure that you want to get ResultSet instead of an array?

romkatsu: It seems logical to me

src/ResultSet.php
```diff
@@ -18,18 +18,23 @@ final class ResultSet implements IteratorAggregate
      * @var Result[]
      */
     private array $results = [];
+    private bool $isValid = true;
+
+    public function addResult(string $attribute, Result $result): void
+    {
+        if (!$result->isValid()) {
+            $this->isValid = false;
```
xepozz, line 26: Please move it lower to 37 line

src/ResultSet.php
xepozz: Why not just flatten array? It will be already invalid so you need only attributes and their results.

xepozz (changes requested):

src/ResultSet.php
romkatsu: This won't work if the first result is error

src/ResultSet.php
romkatsu: fixed

xepozz (approved):

src/ResultSet.php
```diff
@@ -48,4 +53,21 @@ public function getIterator(): ArrayIterator
     {
         return new ArrayIterator($this->results);
     }
+
+    public function getErrors(): array
```
andrew-demb, line 57: Can you describe in phpdoc `@return` tag explicitly specify what items user can expect to be presented in return result?

samdark (approved):

src/ResultSet.php
```diff
@@ -48,4 +53,21 @@ public function getIterator(): ArrayIterator
...
+    {
+        $errors = [];
+        foreach ($this->results as $attribute => $result) {
+            if (!$result->isValid()) {
+                $errors[$attribute] = $result;
```
samdark, line 62: Did you mean to collect errors?

```suggestion
                $errors[$attribute] = $result->getErrors();
```

src/ResultSet.php
romkatsu: done

src/ResultSet.php
romkatsu: done

samdark:
👍

src/ResultSet.php
andrew-demb: As I can see in test (see below), the type should be an `string[][]` (array of arrays of strings), or for psalm `array<string, list<string>>`, not an `string[]`
```php
$this->assertSame(
            ['attribute1' => ['error1', 'error2']],
            $resultSet->getErrors()
);
```

PR #106, opened 2021-01-23
Make abstraction over Translator

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,8 +30,7 @@
         "yiisoft/arrays": "dev-master",
         "yiisoft/friendly-exception": "^1.0",
         "yiisoft/network-utilities": "dev-master",
-        "yiisoft/strings": "^1.0",
-        "yiisoft/translator": "dev-master"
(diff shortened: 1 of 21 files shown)


samdark (approved):

samdark:
Good one but eventually it will lead to https://github.com/yiisoft/validator/pull/99 if we'll remove factories.

xepozz:
> Good one but eventually it will lead to #99 if we'll remove factories.

The difference is that we don't need an interlayer in the form of ErrorMessage; there is no need to throw a formatter into methods, making the API terrible. The formatter can be used as a translator now: simple and convenient.

And yes, `ValidatorFactory` will be removed in #97.

samdark:
https://github.com/yiisoft/validator/pull/97#discussion_r563184026

samdark:
OK. Factories will be removed.

PR #102, opened 2021-01-17
Add validation context

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | -

Static analyse will be pass after release next version of `yiisoft/strings`.

PR #108, opened 2021-01-24
Add optional rule

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -

Based on https://github.com/yiisoft/validator/pull/102


samdark:
The concept of making everything required while having optional rule doesn't fit common forms pattern which is "everything is optional unless required". The pattern is used in [HTML5 client-side validation](https://developer.mozilla.org/en-US/docs/Learn/Forms/Form_validation#using_built-in_form_validation). Going against it would cause pain.

arogachev:
What's the current status of this PR?

arogachev:
Is it somehow related to #132?

samdark:
Related https://github.com/yiisoft/validator/issues/213

Issue #101, opened 2021-01-14
Ability to validate any data


samdark:
It's not good that we require users to implement `DataSetInterface`. Validator should be able to work with any data and should not require and interface.

samdark:
1. Remove typing.
2. If an object implements DataSetInterface - use it.
3. If it is an array or arrayaccess - use it.
4. Else throw exception.

PR #116, opened 2021-02-08
[PoC] Make to able to validate everything

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | https://github.com/yiisoft/validator/issues/101

### DIFF
--- a/src/DataSet/ScalarDataSet.php
+++ b/src/DataSet/ScalarDataSet.php
@@ -6,7 +6,7 @@ namespace Yiisoft\Validator\DataSet;
 
 use Yiisoft\Validator\DataSetInterface;
 
-final class SingleData implements DataSetInterface
+final class ScalarDataSet implements DataSetInterface
 {
     private $value;
 
--- a/src/Rule/CompareTo.php
+++ b/src/Rule/CompareTo.php
@@ -129,10 +129,6 @@ class CompareTo extends Rule
     {
         $result = new Result();
 
-        if ($this->compareValue === null) {
-            throw new \RuntimeException('CompareValidator::compareValue must be set.');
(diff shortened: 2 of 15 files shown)


src/DataSet/ScalarDataSet.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator\DataSet;
+
+use Yiisoft\Validator\DataSetInterface;
+
+final class ScalarDataSet implements DataSetInterface
```
samdark, line 9: What is the use case for this one? It looks a bit out of place.

src/Rule/CompareTo.php
```diff
@@ -129,10 +129,6 @@ protected function validateValue($value, ValidationContext $context = null): Res
     {
         $result = new Result();
 
-        if ($this->compareValue === null) {
```
samdark, line 132: Why?

xepozz: Because `null` is an allowable value too.

src/DataSet/ScalarDataSet.php
xepozz: [It's used for strings, numbers and other scalar types](https://github.com/yiisoft/validator/pull/116/files/f26c7594b8275bb1002476dd6a5425f4628bba12#diff-9dad0f0dabc730596b3399e6fe25aed903dba8bc7836fe778fce387cf9f3e111R62)

samdark (approved):

Issue #120, opened 2021-03-02
Move check exists function "idn_to_ascii" from constructor of the url rule to method enableIDN


vjik:
`enableIDN` **property** is always false in the `Url` rule constructor, so the check for the `idn_to_ascii` function existence will never be performed. We should move this check to the `enableIDN()` **method**.

PR #122, opened 2021-03-02
move checking existance of idn_to_ascii function

kriptonuz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #120

### DIFF
--- a/src/Rule/Url.php
+++ b/src/Rule/Url.php
@@ -9,7 +9,6 @@ use Yiisoft\Validator\Result;
 use Yiisoft\Validator\Rule;
 use Yiisoft\Validator\ValidationContext;
 
-use function function_exists;
 use function is_string;
 use function strlen;
 
--- a/src/Rule/Url.php
+++ b/src/Rule/Url.php
@@ -44,13 +44,6 @@ class Url extends Rule
 
     private string $message = 'This value is not a valid URL.';
 
-    public function __construct()
-    {
-        if ($this->enableIDN && !function_exists('idn_to_ascii')) {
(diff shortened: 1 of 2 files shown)


vjik (approved):

src/Rule/Url.php
```diff
@@ -108,6 +101,11 @@ public function pattern(string $pattern): self
     public function enableIDN(): self
     {
         $new = clone $this;
+
+        if (!function_exists('idn_to_ascii')) {
+            throw new \RuntimeException('In order to use IDN validation intl extension must be installed and enabled.');
+        }
```
viktorprogger, line 107:
```suggestion
        if (!function_exists('idn_to_ascii')) {
            throw new \RuntimeException('In order to use IDN validation intl extension must be installed and enabled.');
        }

        $new = clone $this;
```
There is no need to clone the instance if `idn_to_ascii` doesn't exist.

viktorprogger (changes requested):
This behavior should be checked within a test. Here is an example [how to mock the `function_exists` function](https://stackoverflow.com/questions/32007032/is-it-possible-to-mock-the-non-existence-of-a-function-in-phpunit#answer-34386422).

samdark (approved):
Good change but tests should be added for it.

tests/Rule/UrlTest.php
```diff
@@ -13,6 +25,13 @@
  */
 class UrlTest extends TestCase
 {
+    public static bool $idnFunctionException = false;
```
samdark, line 28: Why do we need it if it's not used anywhere?

samdark:
👍

Issue #121, opened 2021-03-02
TypeError on incorrect url validation with IDN enabled


viktorprogger:
### What steps will reproduce the problem?
`(new Url())->enableIDN()->validate('');`

### What is the expected result?
Return a result object with an error description

### What do you get instead?
> Yiisoft\Validator\Rule\Url::convertIdn(): Return value must be of type string, bool returned at /var/www/vendor/yiisoft/validator/src/Rule/Url.php:82

### Additional info

| Q                | A
| ---------------- | ---
| Version          | dev-master
| PHP version      | 8.0.1
| Operating system | Alpine linux

PR #123, opened 2021-03-06
Fix #121TypeError on incorrect url validation with IDN enabled

DizhovskyyD:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #121

### DIFF
--- a/src/Rule/Url.php
+++ b/src/Rule/Url.php
@@ -70,7 +70,7 @@ class Url extends Rule
         return $result === false ? '' : $result;
     }
 
-    private function convertIdn($value): string
+    private function convertIdn(string $value): string
     {
         if (strpos($value, '://') === false) {
             return $this->idnToAscii($value);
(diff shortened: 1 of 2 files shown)


viktorprogger (approved):

terabytesoftw (approved):

src/Rule/Url.php
```diff
@@ -63,9 +63,11 @@ protected function validateValue($value, ValidationContext $context = null): Res
         return $result;
     }
 
-    private function idnToAscii($idn)
+    private function idnToAscii($idn): string
```
roxblnfk, line 66:
```suggestion
    private function idnToAscii(string $idn): string
```
Should $id be string?

roxblnfk (approved):

src/Rule/Url.php
vjik: Also add type to parameter of method `convertIdn()`.

rustamwin (approved):

samdark (approved):

samdark:
Thanks.

Issue #118, opened 2021-02-26
False positive validation


kriptonuz:
Number rule gives false positive on boolean values
### What steps will reproduce the problem?
```
$rule = new \Yiisoft\Validator\Rule\Number();
$result = $rule->validate(false);
var_dump($result->isValid());
$result = $rule->validate(true);
var_dump($result->isValid());
```
### What is the expected result?
```
false
false
```
### What do you get instead?
```
true
true
```
### Additional info

| Q                | A
| ---------------- | ---
| Version          | 1.0.?
| PHP version      | N/A
| Operating system | N/A

rustamwin:
Boolean values will be converted to integers during validation.

kriptonuz:
Is it normal? For example if I write rule (new Number())->integer(), I will expect integer not boolean values

samdark:
That is tricky. If we deal with standard browser form requests, everything comes as a string from these. JSON-bodies are a bit better but still not ideal. Therefore, type-casting makes sense for input. From another standpoint, using validators standalone is different and sticter comparison is expected.

kriptonuz:
Now I'm using yiisoft/validator for REST API, and I'm validating data came as Json. There are several use cases like this

samdark:
Yes, definitely. Need to think about it.

viktorprogger:
What about implementing a `strictTypes` configuration parameter for the rule?

samdark:
Yeah. That may help. Parameter could be just `strict()`.

PR #124, opened 2021-03-10
Add strict mode to number rule

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️/❌
| Breaks BC?    | ❌
| Fixed issues  | #118

### DIFF
--- a/src/Rule/Number.php
+++ b/src/Rule/Number.php
@@ -23,10 +23,6 @@ class Number extends Rule
      */
     private bool $asInteger = false;
     /**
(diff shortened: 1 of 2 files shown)


vjik:
Maybe don't add option "strict", but make the number rule is always strict?
Has use cases, when number rule should accept boolean values?

samdark:
I don't have any. @rustamwin?

rustamwin:
I don't have too.

vjik (approved):

samdark:
👍

Issue #113, opened 2021-02-02
Use Injector in the Callback rule


viktorprogger:
This will give the ability to use callables with dependencies as validation rules.
One more question here: should the injector be optional to support projects without DI container?

samdark:
The bad thing is that it will make the package dependent on yiisoft/injector.

viktorprogger:
This dependency may be optional

vjik:
Callback rule almost always used into something class, where we can get dependencies from constructor and use their in rule via `$this->`. I think add injector in this rule is overhead.

samdark:
See https://github.com/yiisoft/validator/pull/127#issuecomment-846532958

PR #127, opened 2021-04-04
Using injector in callable rule

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #113

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,9 @@
         "phpunit/phpunit": "^9.5",
         "roave/infection-static-analysis-plugin": "^1.7",
         "spatie/phpunit-watcher": "^1.23",
-        "vimeo/psalm": "^4.3"
+        "vimeo/psalm": "^4.3",
+        "yiisoft/injector": "^1.0",
+        "yiisoft/test-support": "^1.3"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,7 @@
         "ext-mbstring": "*",
         "yiisoft/arrays": "^1.0",
         "yiisoft/friendly-exception": "^1.0",
-        "yiisoft/network-utilities": "dev-master",
+        "yiisoft/network-utilities": "^1.0",
         "yiisoft/strings": "^2.0"
     },
     "require-dev": {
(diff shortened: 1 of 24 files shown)


src/Exception/CallbackRuleException.php
```diff
@@ -10,9 +10,9 @@
 use function get_class;
 use function is_object;
 
-class CallbackRuleException extends \Exception implements FriendlyExceptionInterface
+final class CallbackRuleException extends \Exception implements FriendlyExceptionInterface
 {
-    public function __construct($result, $code = 0, Throwable $previous = null)
+    public function __construct($result, int $code = 0, Throwable $previous = null)
```
root-aza, line 15: ?Throwable $previous = null

composer.json
```diff
@@ -36,7 +36,9 @@
         "phpunit/phpunit": "^9.5",
         "roave/infection-static-analysis-plugin": "^1.7",
         "spatie/phpunit-watcher": "^1.23",
-        "vimeo/psalm": "^4.3"
+        "vimeo/psalm": "^4.3",
+        "yiisoft/injector": "^1.0",
```
samdark, line 40: This one should be in `require`, not `require-dev`.

src/Rule/Callback.php
```diff
@@ -39,4 +43,11 @@ protected function validateValue($value, ValidationContext $context = null): Res
         }
         return $result;
     }
+
+    public function withInjector(Injector $injector): self
```
samdark, line 47: Is there a way to avoid manually passing injector? It's not too bad. Just asking.

samdark (changes requested):

src/Rule/Callback.php
xepozz: I did some research I got that we can have one more variant when Injector will be passed to Rule through Validator. 
It looks stranger than manual passing.

composer.json
xepozz: If you don't need Injector you don't need to install it.

samdark:
Discussed it. Decided not to have it because if you need to pass injector it's better to pass dependecies explicitly.

samdark:
Applied exception improvemnts 👍

Issue #133, opened 2021-05-15
Create rules via ::rule()


vjik:
This is more comfortable. Rules almost always uses with fluent interface.

`(new Number())->integer()`

vs

`Number::rule()->integer()`

samdark:
I like the syntax 👍

PR #134, opened 2021-05-29
Create rules via ::rule()

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #133 

Adopt PRs:
https://github.com/yiisoft/app-api/pull/76
https://github.com/yiisoft/demo/pull/308
https://github.com/yiisoft/demo-api/pull/36
https://github.com/yiisoft/request-model/pull/25
https://github.com/yiisoft/yii-debug-api/pull/32
https://github.com/yiisoft/yii-web/pull/364

### DIFF
--- a/src/Rule.php
+++ b/src/Rule.php
@@ -20,6 +20,30 @@ abstract class Rule implements RuleInterface, ParametrizedRuleInterface, Formatt
      */
     private $when = null;
 
+    /**
+     * To create an instance of a rule class, the static factory method `rule()` must be implemented
+     * in the final class.
+     *
+     * Simple example:
+     *
+     * ```php
+     * public static function rule(): self
+     * {
+     *     return new self();
+     * }
+     * ```
+     *
+     * Example with parameters:
+     *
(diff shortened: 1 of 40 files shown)


src/Rule.php
```diff
@@ -20,6 +20,34 @@ abstract class Rule implements RuleInterface, ParametrizedRuleInterface, Formatt
...
+     * ```php
+     * public static function rule(int $count): self
+     * {
+     *     $rule = new self();
+     *     $rule->count = $count;
+     *     return $rule;
+     * }
+     * ```
+     */
+    final protected function __construct()
```
samdark, line 47: Should it be private?

samdark (approved):

src/Rule.php
vjik: No, in child classes should be available constructor. For example, see rule `AtLeast`.

PR #137, opened 2021-07-22
Add mime validator

iamsaint:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Rule/Mime.php
+++ b/src/Rule/Mime.php
@@ -1,0 +1,54 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator\Rule;
+
+use Psr\Http\Message\UploadedFileInterface;
+use Yiisoft\Validator\Result;
+use Yiisoft\Validator\Rule;
+use Yiisoft\Validator\ValidationContext;
+
+/**
+ * MimeValidator validates that the attribute value is a valid mime type
+ */
+class Mime extends Rule
+{
+    private string $message = 'Incorrect mime type';
+
+    private array $mimeTypes;
+
+    public static function rule(array $mimeTypes): self
+    {
+        $instance = new self();
+        $instance->mimeTypes = $mimeTypes;
+        return $instance;
+    }
+
+    protected function validateValue($value, ValidationContext $context = null): Result
+    {
+        $result = new Result();
+
+        $mimeType = $value;
+
+        if ($value instanceof UploadedFileInterface) {
+            $mimeType = $value->getClientMediaType();
+        }
+
+        if (!in_array($mimeType, $this->mimeTypes, true)) {
+            $result->addError($this->formatMessage($this->message));
+        }
+
+        return $result;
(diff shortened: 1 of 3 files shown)


src/Rule/Mime.php
```diff
@@ -0,0 +1,54 @@
...
+
+use Psr\Http\Message\UploadedFileInterface;
+use Yiisoft\Validator\Result;
+use Yiisoft\Validator\Rule;
+use Yiisoft\Validator\ValidationContext;
+
+/**
+ * MimeValidator validates that the attribute value is a valid mime type
+ */
+class Mime extends Rule
```
vjik, line 15:
```suggestion
final class MimeType extends Rule
```

src/Rule/Mime.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    private string $message = 'Incorrect mime type';
```
vjik, line 17:
```suggestion
    private string $message = 'Incorrect mime type.';
```

src/Rule/Mime.php
```diff
@@ -0,0 +1,54 @@
...
+
+    protected function validateValue($value, ValidationContext $context = null): Result
+    {
+        $result = new Result();
+
+        $mimeType = $value;
+
+        if ($value instanceof UploadedFileInterface) {
+            $mimeType = $value->getClientMediaType();
+        }
```
vjik, line 36:
```suggestion
        $mimeType = $value instanceof UploadedFileInterface
			? $value->getClientMediaType()
			: $value;
```

src/Rule/Mime.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    private string $message = 'Incorrect mime type';
+
+    private array $mimeTypes;
+
+    public static function rule(array $mimeTypes): self
+    {
+        $instance = new self();
+        $instance->mimeTypes = $mimeTypes;
+        return $instance;
```
vjik, line 25:
```suggestion
        $rule = new self();
        $rule->mimeTypes = $mimeTypes;
        return $rule;
```

tests/Rule/MimeTest.php
```diff
@@ -0,0 +1,43 @@
...
+use Yiisoft\Validator\Rule\Mime;
+
+/**
+ * @group validators
+ */
+class MimeTest extends TestCase
+{
+    public function testValidateString()
+    {
+        $val = Mime::rule(['image/jpeg']);
```
vjik, line 18:
```suggestion
        $value = Mime::rule(['image/jpeg']);
```

Don't use trim name, see https://github.com/yiisoft/docs/blob/master/010-code-style.md#names

vjik:
For string values we can use validator `InRange`.

I'm not sure if it's worth making a separate validator for the mime type. Maybe need create validator `File` with more features.

PR #148, opened 2022-01-15
Fix message formater `HasLength::class`.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


src/Rule/HasLength.php
```diff
@@ -39,11 +39,13 @@ final class HasLength extends Rule
     /**
      * @var string user-defined error message used when the length of the value is smaller than {@see $min}.
      */
-    private string $tooShortMessage = 'This value should contain at least {min, number} {min, plural, one{character} other{characters}}.';
```
samdark, line 42: Is there a reason why `min` plural formatting was removed in this an other cases?

terabytesoftw: https://github.com/yiisoft/validator/blob/b2276a349191c00e9e829583712c546a10edf463/src/Formatter.php#L11 The formatter currently does not accept this format, getting the message without formatting.

terabytesoftw: image

xepozz: How about to just improve Formatter instead?

terabytesoftw: > How about to just improve Formatter instead?

Yes it's a good idea, but the formatter should be general, it should be a separate package

samdark: We should not remove plurals. It's better to implenent basic fallback for English language in https://github.com/yiisoft/validator/blob/master/src/Formatter.php.

rustamwin: Use https://github.com/yiisoft/translator-formatter-intl/blob/master/src/IntlMessageFormatter.php instead.

samdark: @rustamwin nope, can't do that because that would create a hard dependency.

terabytesoftw: @samdark  and how will we add support for validator translations ?

terabytesoftw: if the idea is that translation support is outside the package it should be added to the `README.md`.

rustamwin: @samdark Then `MessageFormatter` from intl should be used in https://github.com/yiisoft/validator/blob/master/src/Formatter.php?

samdark: Alternatively we can use https://github.com/yiisoft/translator/blob/master/src/MessageFormatterInterface.php and require it being specified. That's even better.

terabytesoftw: > Alternatively we can use https://github.com/yiisoft/translator/blob/master/src/MessageFormatterInterface.php and require it being specified. That's even better.

And why not use the package in https://github.com/yiisoft/validator/blob/master/src/Formatter.php at the end will be the same dependencies?

samdark: I am suggesting exactly that.

samdark (changes requested):

Issue #147, opened 2022-01-03
Nested validation display errors 


stanislavdev1993:
```php
$request = [
    'author' => [
        'name' => '',
        'age' => '',
    ],
];
 
$rule = Nested::rule([
    'age' => [
        Number::rule(),
    ],
    'name' => [
        HasLength::rule()
    ],
]);
```
 
Got errors

```php
$errors = ['author': ['Value cannot be blank.', 'Value cannot be blank.']]
```
 
Better way to show in what nested attribute got errors

```php
$errors = ['author': ['name': ['Value cannot be blank.'], 'age': ['Value cannot be blank.']]]
```

xepozz:
I think that proposed solution can be improved due to the possibility of unexpected result format.
I have modified your snippet and made an example:
```php
$request = [
    'id' => 1,
    'post' => [
        'author' => [
            'identity' => [
                'name' => '',
                'age' => '',
            ]
        ],
    ],
];
 
$rule = Nested::rule([
    'author.identity.age' => [
        Number()->required(),
    ],
    'author.identity.name' => [
        HasLength::rule()->required(),
    ],
]);
```
The errors will be uncontrollable tree
```php
$result = [
    'post' => [
        'author' => [
            'identity' => [
                'name' => ['Value cannot be blank.'], 
                'age' => ['Value cannot be blank.'],
            ],
        ],
    ],
];
```

One way how to get the error messages is to check that array keys are in int format.

----

If we concatenate nested attributes with `.` symbol, then we need to escape or replace that symbol or even make it reservable.

In conclusion I want to say that we need to think about other formats and choose the best one before implementing.

arogachev:
@stanislavdev1993 Where you got this from?

```php
Number()->required(),
```

It's not in API. Or it's sort of draft / pseudocode?

```php
$errors = ['author': ['Value cannot be blank.', 'Value cannot be blank.']]
```

For me output is flat array. Could you show the rest of the code?

stanislavdev1993:
@arogachev It was mistake. I fixed this in comment.

arogachev:
Got it, no problem. But what about chained `required` call?

stanislavdev1993:
@arogachev Done. It just a sample. issue with errors diplaying in nested rule.

Issue #135, opened 2021-06-17
Validate nested data


andrew-svirin:
Hi,
Is any example how to validate complex data?
Like:
```php
[
 'name' => 'Name',
 'address' => [
   'city' => 'City', 
 ],
]
```

samdark:
https://github.com/yiisoft/validator/pull/98

andrew-svirin:
Perfect!

andrew-svirin:
And this option is also possible?
```php
[
 'users' => [
   [
     'name' => 'Name1',
   ],
   [
     'name' => 'Name2',
   ]
 ]
]
```

samdark:
I don't remember. Try it.

@xepozz maybe you remember if it was implemented or just planned to be implemented?

xepozz:
I don't sure that it's possible as `NestedRule`, but you can try to combine it with `Each` or another rules.

samdark:
@andrew-svirin please post about your results. Will help us.

andrew-svirin:
@samdark Okay. I will do soon.

andrew-svirin:
First of all - Thank you!

Q1:
Every time worked with complex DTO I faced problem in validation:
For example incoming request:
```json
{
 "username": "Nick",
 "email": "test@email",
 "purchase" : [
    {
      "product_id": "111-111-abc",
      "amount" : 111
    }
    ...
 ]
...
}
``` 
In example is demonstrated that request can hold multiple entities those should be saved in transaction.
Question is:
1. Should I combine multiple Validators to validate DTO? Does Yii support reuse validators?
2. Should I duplicate validation rules?
Could you suggest the correct approach when using Yii3.

Q2:
Will be road map how to migrate business logic from Yii2 to Yii3? How to write validation for js framework? What js framework will recommend Yii3?

And also I collected some other tools that could be common problems when working with framework:
1. phpstorm-plugin for Yii3. This is a big plus to any framework.
2. Flexible event-listener configuration to realise relations between modules. And is possible to have cached configuration-per-controller?
3. Flexible declaration relations between entities that can be used for building composite data response?
4. Would Yii3 can make some services private for module?

xepozz:
> Should I combine multiple Validators to validate DTO?
Should I duplicate validation rules?

As you wish.

> Does Yii support reuse validators?

[Yes, it is](https://github.com/yiisoft/validator/#creating-your-own-validation-rules).


> How to write validation for js framework?

[...]

PR #151, opened 2022-01-24
Validate nested data, add error details

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #135, #147 

TODO list:

- [x] Check forms and other repositories using validator for BC break.
- [x] Allow to use flat array as input. Can be done in a separate issue though. Or maybe this should be implemented at another level (forms, etc. - need to discuss). Moved to #154.


src/Rule/Each.php
```diff
@@ -37,20 +37,27 @@ protected function validateValue($value, ValidationContext $context = null): Res
...
-                    $result->addError(
-                        $this->formatMessage(
-                            $this->message,
-                            [
-                                'error' => $error,
-                                'value' => $item,
-                            ]
-                        )
-                    );
+            if ($itemResult->isValid() === true) {
```
vjik, line 42:
```suggestion
            if ($itemResult->isValid()) {
```

src/Rule/Each.php
```diff
@@ -37,20 +37,27 @@ protected function validateValue($value, ValidationContext $context = null): Res
...
+                continue;
+            }
+
+            foreach ($itemResult->getErrors() as $key => $error) {
+                if (!is_array($item)) {
+                    $errorKey = $index;
+                    $formatMessage = true;
+                } else {
+                    $errorKey = $index . Result::ERROR_KEY_SEPARATOR . $key;
+                    $formatMessage = false;
```
vjik, line 52: Why not format message?

README.md
```diff
@@ -17,6 +17,33 @@ The package provides data validation capabilities.
...
+        * [Setting up your own formatter](#setting-up-your-own-formatter)
+    * [Testing](#testing)
+        * [Unit testing](#unit-testing)
+        * [Mutation testing](#mutation-testing)
+        * [Static analysis](#static-analysis)
+    * [License](#license)
+    * [Support the project](#support-the-project)
+    * [Follow updates](#follow-updates)
+
+Created by [gh-md-toc](https://github.com/ekalinin/github-markdown-toc)
```
samdark, line 45: I don't think we need it in README.

src/Rules.php
```diff
@@ -50,8 +50,8 @@ public function validate($value, ValidationContext $context = null): Result
             $ruleResult = $rule->validate($value, $context);
             if ($ruleResult->isValid() === false) {
                 $context->setParameter(self::PARAMETER_PREVIOUS_RULES_ERRORED, true);
-                foreach ($ruleResult->getErrors() as $message) {
-                    $compoundResult->addError($message);
+                foreach ($ruleResult->getErrors() as $key => $message) {
+                    $compoundResult->addError($message, $key);
```
vjik, line 54:
```suggestion
                    $compoundResult->addError($message, is_string($key) ? $key : null);
```

Seems, for number keys errors from diefferent rules can be overrided.

src/Rule/Nested.php
```diff
@@ -169,4 +168,36 @@ private function fetchOptions(array $rules): array
...
+
+        $aggregatedRuleArr = $aggregatedRule->asArray();
+        if (ArrayHelper::getValueByPath($aggregatedRuleArr, '0.0') !== 'each') {
+            return false;
+        }
+
+        return !(ArrayHelper::getValueByPath($aggregatedRuleArr, '0.1.0') === 'nested');
+    }
+
+    private static function isPositiveInteger(string $str): bool
```
vjik, line 195: Maybe add this method to Yii Strings: https://github.com/yiisoft/strings/blob/master/src/NumericHelper.php ?

README.md
```diff
@@ -142,6 +169,81 @@ To change this behavior use `skipOnEmpty(true)`:
 Number::rule()->integer()->max(100)->skipOnEmpty(true)
 ```
 
+#### Nested and related data
+
```
samdark, line 173:
```suggestion
#### Nested and related data

In many cases there is a need to validate data consisting of many similar structures. There is a `Nested` rule for this purpose:
```

src/Rule/Nested.php
```diff
@@ -169,4 +168,36 @@ private function fetchOptions(array $rules): array
...
+        if (!self::canConcatenateErrorKey($aggregatedRule, $key)) {
+            return $valuePath;
+        }
+
+        return $valuePath . Result::ERROR_KEY_SEPARATOR . $key;
+    }
+
+    private static function canConcatenateErrorKey(Rules $aggregatedRule, string $key): bool
+    {
+        if (!self::isPositiveInteger($key)) {
```
vjik, line 183: Seems, we can check is_int($key) in `foreach ($itemResult->getErrors() as $key => $error) {` above.

src/Rule/Nested.php
```diff
@@ -169,4 +168,36 @@ private function fetchOptions(array $rules): array
...
+            return true;
+        }
+
+        $aggregatedRuleArr = $aggregatedRule->asArray();
```
vjik, line 187:
```suggestion
        $aggregatedRuleArray = $aggregatedRule->asArray();
```

vjik (changes requested):

README.md
```diff
@@ -142,6 +169,81 @@ To change this behavior use `skipOnEmpty(true)`:
...
+$rule = Nested::rule([
+    'author' => Nested::rule([
+        'name' => [HasLength::rule()->min(3)],
+        'age' => [Number::rule()->min(18)],
+    )];
+]);
+$errors = $rule->validate($data)->getErrors();
+```
+
+A more complex real-life example using arrays.
```
samdark, line 190:
```suggestion
A more complex real-life example is a chart that is made of points. This data is represented as an arrays:
```

README.md
```diff
@@ -142,6 +169,81 @@ To change this behavior use `skipOnEmpty(true)`:
...
+                    ])),
+                ])->skipOnError(false),
+            ])),
+        ])->skipOnError(false),
+    ])),
+]);
+$errors = $rule->validate($data)->getErrors();
+```
+
+The contents of the errors will be (omitted for brevity):
```
samdark, line 238:
```suggestion
The contents of the errors will be:
```

README.md
```diff
@@ -142,6 +169,81 @@ To change this behavior use `skipOnEmpty(true)`:
...
+
+```php
+$errors = [
+    'charts.0.dots.0.coordinates.x' => 'Value must be no less than -10.',
+    'charts.0.dots.0.rgb.0' => 'Value must be no less than 0. -1 given.',
+];
```
samdark, line 244:
```suggestion
    'charts.0.dots.0.rgb.0' => 'Value must be no less than 0. -1 given.',
    // ...
];
```

src/Result.php
```diff
@@ -6,8 +6,10 @@
 
 final class Result
 {
+    public const ERROR_KEY_SEPARATOR = '.';
```
samdark, line 9: Using `.` directly without constant looks way better to me. There's not that many chances to mistype with `.` as with longer strings.

src/Result.php
```diff
@@ -16,13 +18,17 @@ public function isValid(): bool
         return $this->errors === [];
     }
 
-    public function addError(string $message): void
+    public function addError(string $message, $key = null): void
```
samdark, line 21: The method now needs phpdoc since it's not clear what `$key` is from method signature.

src/Rule/Nested.php
```diff
@@ -169,4 +168,36 @@ private function fetchOptions(array $rules): array
...
+        if (ArrayHelper::getValueByPath($aggregatedRuleArr, '0.0') !== 'each') {
+            return false;
+        }
+
+        return !(ArrayHelper::getValueByPath($aggregatedRuleArr, '0.1.0') === 'nested');
+    }
+
+    private static function isPositiveInteger(string $str): bool
+    {
+        if (!preg_match('/^\d$/', $str)) {
```
samdark, line 197: Is this check here to exclude scientific notation?

samdark (changes requested):

src/Rule/Nested.php
```diff
@@ -169,4 +168,36 @@ private function fetchOptions(array $rules): array
...
+    }
+
+    private static function canConcatenateErrorKey(Rules $aggregatedRule, string $key): bool
+    {
+        if (!self::isPositiveInteger($key)) {
+            return true;
+        }
+
+        $aggregatedRuleArr = $aggregatedRule->asArray();
+        if (ArrayHelper::getValueByPath($aggregatedRuleArr, '0.0') !== 'each') {
```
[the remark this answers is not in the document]
arogachev: That's how it was before. Changed it because it's a bit more compact and readable that way. Not sure performance gain will be big enough here (without benchmarks). As always, performance vs readability ("design first, optimize later" principle). But OK, I'll change it back.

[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Rule/Nested.php
```diff
@@ -169,4 +168,36 @@ private function fetchOptions(array $rules): array
...
+            return false;
+        }
+
+        return !(ArrayHelper::getValueByPath($aggregatedRuleArr, '0.1.0') === 'nested');
```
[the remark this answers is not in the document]
arogachev: https://github.com/yiisoft/validator/pull/151#discussion_r792507450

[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Result.php
```diff
@@ -6,8 +6,10 @@
 
 final class Result
 {
+    public const ERROR_KEY_SEPARATOR = '.';
+
     /**
-     * @psalm-var list<string>
+     * @var array<mixed, string>
```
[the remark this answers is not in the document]
arogachev: Got these docs a bit wrong. https://psalm.dev/docs/annotating_code/type_syntax/array_types/

Corrected (including the entry in `getErrors()` method's docblock). I also replaced `mixed` with more specific `int|string`. Or did you mean replace the whole line with `@var string[]? The idea was to document keys as well.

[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Result.php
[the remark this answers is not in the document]
arogachev: Left like that because `$key` can also be `int`. But passed `$key` is always `string` though, you are right. Corrected.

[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Result.php
```diff
@@ -16,13 +18,17 @@ public function isValid(): bool
         return $this->errors === [];
     }
 
-    public function addError(string $message): void
+    public function addError(string $message, $key = null): void
     {
-        $this->errors[] = $message;
+        if ($key) {
```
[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Rule/Each.php
arogachev: But it was strictly checked for `false` before, I just refactored it to continue early to reduce indentation level and improve readability (following best practices). A bit confused. But it seems to be safe here because of using `bool`.

arogachev: :heavy_check_mark:

src/Result.php
```diff
@@ -16,13 +18,17 @@ public function isValid(): bool
...
+            $this->errors[$key] = $message;
+        } else {
+            $this->errors[] = $message;
+        }
     }
 
     /**
-     * @psalm-return list<string>
+     * @return array<mixed, string>
```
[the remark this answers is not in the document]
arogachev: https://github.com/yiisoft/validator/pull/151#discussion_r792520707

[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Result.php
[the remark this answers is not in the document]
arogachev: It actually can be `int` with combinations of `Nested` and `Each`, so I am reverting this back. I can't add `mixed` here because it's only possible since PHP 8. Specifying multiple types is also not possible as far as I remember. At least I can add this to PHPDoc docblock.

src/Rule/Each.php
arogachev: Otherwise you get stuff like that in errors:

```php
'charts.2.dots.1.rgb.2' => "Value must be no greater than 255. 259 given. array given. array given."
```

After this change it's returned as expected:

```php
'charts.2.dots.1.rgb.2' => "Value must be no greater than 255. 259 given."

It's needed for correct handling of `Nested` / `Each` combinations
```

arogachev: :heavy_check_mark:

arogachev: Also was thinking about removing value from error message completely, but it's a regression and less intuitive.

src/Rules.php
arogachev: Sorry. I didn't understand this one. How it can be overridden? Maybe provide an example? Also, in case of adding this, maybe move it inside `addError` method?

src/Rule/Nested.php
arogachev: Yep, sounds like a good idea. Thought about adding this to a more generic helper within the current repository. But indeed, we can add this to `NumericHelper`. Will do later.

src/Rule/Nested.php
arogachev: You mean use different logic for detecting a positive integer? Or move it to the loop? Or both?  Sorry, but in case of the latter, how it will improve the current code? Moving just this check is not enough too. What's the reason of extracting it?

src/Rule/Nested.php
arogachev: :heavy_check_mark:

README.md
arogachev: Sorry, not 100% clear for me.

You mean the whole TOC? Why not? It's getting quite big and becoming harder to navigate. For better sync we can automate it. 

Or you discourage using this specific tool? 

Or mentioning it in `README` (so you mean only the line "Created by `gh-md-toc`..." is redundant)? But I think it's useful - it marks this block as auto generated (we can also add more explicit note) and it contains link to instructions on how to update it.

samdark: We are eventually going to render these at the website like https://www.yiiframework.com/extension/yiisoft/yii2-redis. There we generate TOC automatically:

image


"Created by gh-md-toc" despite being technically useful is very discouraging for the user of the library.

src/Result.php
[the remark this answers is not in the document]
samdark: I think it's fine as is.

src/Result.php
```diff
@@ -16,13 +18,22 @@ public function isValid(): bool
...
+        if ($key !== null) {
+            $this->errors[$key] = $message;
+        } else {
+            $this->errors[] = $message;
+        }
     }
 
     /**
-     * @psalm-return list<string>
+     * @psalm-var list<int|string, string>
```
samdark, line 36:
```suggestion
     * @psalm-param list<int|string, string>
```

README.md
arogachev: > We are eventually going to render these at the website

But it's not ready yet. And I think there are still many users reading docs directly from Github, but I don't have stats to prove it.

Anyway, I got it. Removed the whole block.

arogachev: :heavy_check_mark:

README.md
arogachev: Good addition. But I think I'll correct this a bit. `Nested` is more about relations, while `Each` - about similar structures.

arogachev: :heavy_check_mark:

README.md
arogachev: Thanks. Added with small correction and addition. See https://github.com/yiisoft/validator/pull/151#discussion_r793256574. Also fixed a grammar error.

arogachev: :heavy_check_mark:

src/Result.php
arogachev: It's actually neither `var`, nor `param`. It's `return`. Thank you for noticing. Fixed. Probably was copy pasted from above.

arogachev: :heavy_check_mark:

README.md
arogachev: :heavy_check_mark:

README.md
arogachev: :heavy_check_mark:

src/Result.php
arogachev: If I got this correctly, this was suggested in https://github.com/yiisoft/validator/issues/147#issuecomment-1004499792 by @xepozz. I'm fine with using it without constant too because it's such a common notation in other languages / frameworks.

arogachev: :heavy_check_mark:

src/Result.php
arogachev: Added. See https://github.com/yiisoft/validator/pull/151#discussion_r792535726.

arogachev: :heavy_check_mark:

src/Rule/Nested.php
arogachev: It's needed to additionally exclude strings like `+5`. See https://stackoverflow.com/a/15277806/4323648 (2nd comment).

arogachev: :heavy_check_mark:

src/Result.php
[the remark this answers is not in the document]
arogachev: This actually causes unit tests to fail now. Adjusted for strict checking.

src/Result.php
samdark: I've meant description of what `$key` is for, not just types.

src/Rule/Nested.php
arogachev: https://github.com/yiisoft/strings/pull/78

arogachev: :heavy_check_mark:

arogachev: Also adjusted to not consider 0 as positive integer.

arogachev: Tests are now failing in CI because we need to release https://github.com/yiisoft/strings/pull/78 first. Tested localy with new changes, they are passing.

src/Result.php
arogachev: OK. Added.

src/Rule/Nested.php
arogachev: Refactored. Please check.

src/Rule/Nested.php
arogachev: I simplified and refactored implementation and it's no longer needed here. Still can be useful in strings package.

src/Rule/Nested.php
arogachev: :heavy_check_mark:

src/Result.php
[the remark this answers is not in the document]
vjik: `@psalm-var array<int|string, string>` === `@var string[]`

Because array key can be int or string only.

vjik (commented):
This test fail, but should be pass.

```php
public function testGetErrorsWithSeveralErrors(): void
{
	$rules = new Rules(
		[
			Callback::rule(
				static function ($value): Result {
					$result = new Result();
					$result->addError('e1');
					$result->addError('e2');
					$result->addError('e3');
					return $result;
				}
			),
			Callback::rule(
				static function ($value): Result {
					$result = new Result();
					$result->addError('e4');
					$result->addError('e5');
					$result->addError('e6');
					return $result;
				}
			)->skipOnError(false),
		]
	);

	$result = $rules->validate('hi');

	$this->assertFalse($result->isValid());
	$this->assertSame([
		0 => 'e1',
		1 => 'e2',
		2 => 'e3',
		3 => 'e4',
		4 => 'e5',
		5 => 'e6',
	], $result->getErrors());
}
```

src/Rules.php
```diff
@@ -99,6 +100,16 @@ public function asArray(): array
...
     }
 
+    public function isNestedEach(): bool
+    {
+        $array = $this->asArray();
+        if (ArrayHelper::getValue($array, [0, 0]) !== 'each') {
+            return false;
+        }
+
+        return !(ArrayHelper::getValue($array, [0, 1, 0]) === 'nested');
```
xepozz, line 110:
```suggestion
        return ArrayHelper::getValue($array, [0, 1, 0]) !== 'nested';
```
Why not like this?

arogachev: This was probably suggested by StyleCI at some point. Indeed, can be simplified. Thanks.

arogachev: :heavy_check_mark:

src/Result.php
```diff
@@ -16,13 +16,23 @@ public function isValid(): bool
         return $this->errors === [];
     }
 
-    public function addError(string $message): void
+    /**
+     * @param string $message
```
samdark, line 20:
```suggestion
     * Add an error.
     *
     * @param string $message Error message.
```

samdark (approved):

samdark:
@arogachev let's add the test above from @vjik.

src/Result.php
arogachev: It's literally just a method name. Maybe extend it a bit with details or do not add at all yet? And it will be inconsistent with other missing descriptions, for example, for `isValid()` method, etc.

arogachev: Or maybe there are some other documentation nuances involved here?

src/Rules.php
arogachev: Now I get it. :slightly_smiling_face:

arogachev: :heavy_check_mark:

arogachev:
@vjik Thanks for this. Fixed. Please take a look. Is this behavior OK? Or maybe we should at least update the docs about persistent errors? Tried other fixes, but nested and each tests are failing.

src/Error.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * @psalm-return array<int|string>
+     */
+    public function getValuePath(): ?array
+    {
+        return $this->valuePath;
+    }
+
+    public function getStringValuePath(): string
```
vjik, line 31: We can remove this method. See comments below.

src/Result.php
```diff
@@ -16,16 +18,63 @@ public function isValid(): bool
         return $this->errors === [];
     }
 
-    public function addError(string $message): void
+    /**
+     * @psalm-param list<int|string>|null $valuePath
+     */
+    public function addError(string $message, ?array $valuePath = null): void
```
vjik, line 24:
```suggestion
    public function addError(string $message, array $valuePath = []): void
```

src/Error.php
```diff
@@ -0,0 +1,35 @@
...
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+final class Error
+{
+    private string $message;
+    private ?array $valuePath;
+
+    public function __construct(string $message, ?array $valuePath = null)
```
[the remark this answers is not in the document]
vjik: We don't need to distinguish between "no path" and "empty path".

src/Result.php
```diff
@@ -16,16 +18,63 @@ public function isValid(): bool
...
+
+    /**
+     * @return string[]
+     */
+    public function getErrors(): array
+    {
+        return ArrayHelper::getColumn($this->errors, function ($error) {
+            /** @var Error $error */
+            return $error->getMessage();
+        });
```
vjik, line 45:
```suggestion
        return ArrayHelper::getColumn($this->errors, static fn (Error $error) => $error->getMessage());
```

src/Result.php
```diff
@@ -16,16 +18,63 @@ public function isValid(): bool
...
+        $errors = [];
+        foreach ($this->errors as $error) {
+            $stringValuePath = $error->getStringValuePath();
+            $valuePathCount = $valuePathCountMap[$stringValuePath] ?? 0;
+            $errorValuePath = "$stringValuePath.$valuePathCount";
+
+            ArrayHelper::setValueByPath($errors, $errorValuePath, $error->getMessage());
+            $valuePathCount++;
+            $valuePathCountMap[$stringValuePath] = $valuePathCount;
+        }
```
vjik, line 63:
```suggestion
$errors = [];
        foreach ($this->errors as $error) {
            $valuePath = $error->getValuePath();
            if ($valuePath === []) {
                $errors[0][] = $error->getMessage();
            } else {
                $errorsArray = ArrayHelper::getValue($errors, $valuePath, []);
                $errorsArray[] = $error->getMessage();
                ArrayHelper::setValue($errors, $valuePath, $errorsArray);
            }
        }
```

src/Result.php
```diff
@@ -16,16 +18,63 @@ public function isValid(): bool
...
+    }
+
+    /**
+     * @return array
+     */
+    public function getFlatDetailedErrors(): array
+    {
+        $errors = [];
+        foreach ($this->errors as $error) {
+            $errors[$error->getStringValuePath()][] = $error->getMessage();
```
vjik, line 75:
```suggestion
            $errors[implode('.', $error->getValuePath())][] = $error->getMessage();
```

src/Result.php
```diff
@@ -16,16 +18,63 @@ public function isValid(): bool
...
+        return ArrayHelper::getColumn($this->errors, function ($error) {
+            /** @var Error $error */
+            return $error->getMessage();
+        });
+    }
+
+    /**
+     * @return array
+     */
+    public function getNestedDetailedErrors(): array
```
vjik, line 51:
```suggestion
    public function getNestedErrors(): array
```

src/Result.php
```diff
@@ -16,16 +18,63 @@ public function isValid(): bool
...
-        $this->errors[] = $message;
+        $this->errors[] = new Error($message, $valuePath);
     }
 
     /**
-     * @psalm-return list<string>
+     * @return Error[]
      */
-    public function getErrors(): array
+    public function getErrorObjects(): array
```
vjik, line 32: Need same method in `ResultSet`.

src/Result.php
vjik, line 51: Need same method in `ResultSet`.

src/Result.php
```diff
@@ -16,16 +18,63 @@ public function isValid(): bool
...
+            $valuePathCountMap[$stringValuePath] = $valuePathCount;
+        }
+
+        return $errors;
+    }
+
+    /**
+     * @return array
+     */
+    public function getFlatDetailedErrors(): array
```
vjik, line 71: Need same method in `ResultSet`.

vjik, line 71:
```suggestion
    public function getErrorsIndexedByPath(string $separator = '.'): array
```

src/Result.php
vjik, line 24: Test with collision:

$data = [
  'names' => [
    'god89',
    'greatJonh',
    'vasya77',
  ],
];
$rules = [
  'names' => [
    Each::rule(new Rules([
      MatchRegularExpression::rule('~\d~')->not()->skipOnError(false),
    ])),
    Number::rule()->skipOnError(false),
  ],
];

$nestedResult = Nested::rule($rules)->validate($data);
$this->assertSame([
  'names' => [
    ???
  ],  
], $nestedResult->getNestedDetailedErrors());

What result do we expect?

`names` contain both attributes with errors and errors themselves.

Ideas::

1) Remove errors of `name` itself:

0 => [
  'Value is invalid. god89 given.',
],
2 => [
  'Value is invalid. vasya77 given.',
],

2) Add errors in `names` (bad variant, can be collision with keys):

0 => [
  'Value is invalid. god89 given.',
],
2 => [
  'Value is invalid. vasya77 given.',
],
3 => 'Value must be a number.',

3) Throw exception.

src/Result.php
```diff
@@ -16,16 +18,63 @@ public function isValid(): bool
         return $this->errors === [];
     }
 
-    public function addError(string $message): void
+    /**
+     * @psalm-param list<int|string>|null $valuePath
```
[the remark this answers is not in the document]
vjik: > Are you sure it may include int?

Yes, for example array of fields that validate by `Each` rule.

src/Error.php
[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Result.php
arogachev: :heavy_check_mark:

src/Error.php
```diff
@@ -0,0 +1,35 @@
...
+
+    public function getMessage(): string
+    {
+        return $this->message;
+    }
+
+    /**
+     * @psalm-return array<int|string>
+     */
+    public function getValuePath(): ?array
```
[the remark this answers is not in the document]
arogachev: Yes, see this for example https://github.com/arogachev/validator/blob/fcd149d6542bf722edc1d5a66445306dc6f3fb0c/tests/Rule/NestedTest.php#L237. Or should we cast them to string?

src/Result.php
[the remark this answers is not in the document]
arogachev: https://github.com/yiisoft/validator/pull/151#discussion_r798270000

src/Error.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+final class Error
+{
+    private string $message;
+    private ?array $valuePath;
```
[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Error.php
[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Result.php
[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Rule/Nested.php
```diff
@@ -81,30 +79,35 @@ protected function validateValue($value, ValidationContext $context = null): Res
...
-            if ($itemResult->isValid() === false) {
-                foreach ($itemResult->getErrors() as $error) {
-                    $result->addError($error);
+            $itemResult = $aggregatedRule->validate($validatedValue);
+            if ($itemResult->isValid()) {
+                continue;
+            }
+
+            foreach ($itemResult->getErrorObjects() as $error) {
+                $errorValuePath = is_array($valuePath) ? $valuePath : explode('.', $valuePath);
```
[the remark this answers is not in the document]
arogachev: How can it be? Could you provide an example?

src/Result.php
arogachev: :heavy_check_mark:

src/Rule/Nested.php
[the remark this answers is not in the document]
vjik: > How can it be? Could you provide an example?

When attribute name is number. Try add rules to '0' => ...

src/Result.php
arogachev: :heavy_check_mark:

src/Error.php
arogachev: :heavy_check_mark:

src/Result.php
arogachev: :heavy_check_mark:

src/Result.php
arogachev: :heavy_check_mark:

arogachev: Thanks, added with minor adjustments.

src/Result.php
arogachev: :heavy_check_mark:

src/Rule/Nested.php
arogachev: With new approach (using `Error` objects) there is no check for concatenation anymore, so this is no longer actual.

src/Rule/Nested.php
arogachev: No longer actual too.

src/Rule/Nested.php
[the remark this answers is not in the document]
arogachev: Indeed. Not so evident in case of using `Nested`. Maybe add this to the docs as well? :thinking: By the way I caught something like this during debug at some point and then it was gone. :astonished:

src/Result.php
arogachev: :heavy_check_mark:

src/Result.php
arogachev: :heavy_check_mark:

src/Result.php
arogachev: :heavy_check_mark:

src/Error.php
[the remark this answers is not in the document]
vjik: Cast to string doesn't make sense. For PHP integer and string array keys is same.

src/Result.php
```diff
@@ -16,16 +18,63 @@ public function isValid(): bool
...
-    public function getErrors(): array
+    public function getErrorObjects(): array
     {
         return $this->errors;
     }
+
+    /**
+     * @return string[]
+     */
+    public function getErrors(): array
```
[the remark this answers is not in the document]
arogachev: After discussion in messenger, we decided to do it in the next PR (because of amount of changes and affect on form and demo).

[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Result.php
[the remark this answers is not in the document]
arogachev: We discussed it in messenger, it's OK as is.

src/Error.php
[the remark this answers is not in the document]
arogachev: We discussed it in messenger, it's OK as is.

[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Result.php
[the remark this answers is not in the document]
arogachev: :heavy_check_mark:

src/Result.php
arogachev: After discussion in messenger we decided to postpone tree implementation for the next iteration.

arogachev: :heavy_check_mark:

README.md
```diff
@@ -142,6 +142,87 @@ To change this behavior use `skipOnEmpty(true)`:
...
+]);
+$errors = $rule->validate($data)->getErrors();
+```
+
+The contents of the errors will be:
+
+```php
+$errors = [
+    'charts.0.points.0.coordinates.x' => 'Value must be no less than -10.',
+    'charts.0.points.0.rgb.0' => 'Value must be no less than 0. -1 given.',
```
vjik, line 221:
```suggestion
    'Value must be no less than -10.',
    'Value must be no less than 0. -1 given.',
```

src/Error.php
```diff
@@ -0,0 +1,30 @@
...
+
+    public function getMessage(): string
+    {
+        return $this->message;
+    }
+
+    /**
+     * @psalm-return array<int|string>
+     */
+    public function getValuePath(): ?array
```
vjik, line 26:
```suggestion
    public function getValuePath(): array
```

src/ErrorsReadInterface.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+interface ErrorsReadInterface
```
vjik, line 7: Suggest remove this interfaceand in future prs merge `ResultSet` and `Result` to one class.

src/ResultSet.php
```diff
@@ -21,6 +22,50 @@ final class ResultSet implements IteratorAggregate
...
+    {
+        return $this->getAttributeToErrorsMap(static fn (Result $result) => $result->getErrors());
+    }
+
+    public function getErrorsIndexedByPath(string $separator = '.'): array
+    {
+        return $this->getAttributeToErrorsMap(static fn (Result $result) => $result->getErrorsIndexedByPath($separator));
+    }
+
+    private function getAttributeToErrorsMap(Closure $getErrorsClosure): array
```
vjik, line 57:
```suggestion
    private function errorsMap(Closure $getErrorsClosure): array
```

src/Rule/Each.php
```diff
@@ -37,20 +37,27 @@ protected function validateValue($value, ValidationContext $context = null): Res
...
+            if ($itemResult->isValid()) {
+                continue;
+            }
+
+            foreach ($itemResult->getErrorObjects() as $error) {
+                if (!is_array($item)) {
+                    $errorKey = [$index];
+                    $formatMessage = true;
+                } else {
+                    $errorKey = array_merge([$index], $error->getValuePath());
```
vjik, line 51:
```suggestion
                    $errorKey = [$index, ...$error->getValuePath()];
```

src/Rule/Nested.php
```diff
@@ -81,30 +79,42 @@ protected function validateValue($value, ValidationContext $context = null): Res
...
-            if ($itemResult->isValid() === false) {
-                foreach ($itemResult->getErrors() as $error) {
-                    $result->addError($error);
+            $itemResult = $aggregatedRule->validate($validatedValue);
+            if ($itemResult->isValid()) {
+                continue;
+            }
+
+            foreach ($itemResult->getErrorObjects() as $error) {
+                if (is_array($valuePath)) {
```
vjik, line 105: `$valuePath` always is integer or string. This condition always is false.

tests/ResultSetTest.php
```diff
@@ -40,40 +57,49 @@ public function testErrorsShouldAdd(): void
...
 
-    public function testIsValidReturnFalse(): void
+    public function testGetErrorsIndexedByPath(): void
     {
-        $resultSet = new ResultSet();
-        $resultSet->addResult('attribute1', $this->createErrorResult('error1'));
-
-        $this->assertFalse($resultSet->isValid());
+        $this->assertEquals(
+            ['attribute1' => ['' => ['error1', 'error2']], 'attribute2' => ['path.3' => ['error3']]],
```
vjik, line 82:
```suggestion
            ['attribute1' => ['attribute1' => ['error1', 'error2']], 'attribute2' => ['path.3' => ['error3']]],
```
`ResultSet` known about `attribute1` and must take this into account.

README.md
arogachev: Thanks, but I think in this context it's better to change the called method.

arogachev: :heavy_check_mark:

src/Error.php
arogachev: :heavy_check_mark:

src/ErrorsReadInterface.php
arogachev: :heavy_check_mark:

arogachev: We could remove it in the same future PR too, but OK, removed.

src/ResultSet.php
arogachev: Yes, it's probably fine to use shorter variation here, but I'd leave `get` prefix.

arogachev: :heavy_check_mark:

src/Rule/Each.php
arogachev: :heavy_check_mark:

src/Rule/Nested.php
arogachev: :heavy_check_mark:

src/Rule/Each.php
arogachev: Reverted because of Psalm error. https://github.com/yiisoft/validator/runs/5076273960?check_suite_focus=true What should we do about this?

tests/ResultSetTest.php
arogachev: So with this logic to be consistent for 'attribute2' it should be:

```php
'attribute2' => ['attribute.path.3' => ['error3']`
```

?

src/Rule/Each.php
vjik: Modify `Error` for fix psalm error:

```php
final class Error
{
    private string $message;

    /**
     * @psalm-var list<int|string>
     */
    private array $valuePath;

    /**
     * @psalm-param list<int|string> $valuePath
     */
    public function __construct(string $message, array $valuePath = [])
    {
        $this->message = $message;
        $this->valuePath = $valuePath;
    }

    public function getMessage(): string
    {
        return $this->message;
    }

    /**
     * @psalm-return list<int|string>
     */
    public function getValuePath(): array
    {
        return $this->valuePath;
    }
}
```

tests/ResultSetTest.php
vjik: Sorry. I'm was wrong.

Maybe this should be:

```php
[
    'attribute1' => ['error1', 'error2'],
    'attribute2.path.3' => ['error3'],
],
```

?

arogachev: :heavy_check_mark:

src/Rule/Each.php
arogachev: :heavy_check_mark:

vjik (approved):

samdark (approved):

samdark:
👍

PR #163, opened 2022-02-07
Remove ResultSet

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #161


xepozz:
What's the purpose?

arogachev:
@xepozz We decided that it's redundant after #151, because now every error contains value path and knows about attribute. Initially that was @vjik's suggestion.

src/PostValidationHookInterface.php
```diff
@@ -11,5 +11,8 @@
  */
 interface PostValidationHookInterface extends DataSetInterface
 {
-    public function processValidationResult(ResultSet $resultSet): void;
+    /**
+     * @param Result $result
+     */
```
vjik, line 16:
```suggestion
```

This is redundant.

vjik (approved):

src/PostValidationHookInterface.php
arogachev: Agree. Thanks, fixed. :+1:

arogachev: :heavy_check_mark:

samdark (approved):

samdark:
👍

PR #164, opened 2022-02-07
Rename Rules to RuleSet

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #153

### DIFF
--- a/src/Rule/Each.php
+++ b/src/Rule/Each.php
@@ -25,7 +25,7 @@ final class Each extends Rule
     public static function rule(RuleSet $ruleSet): self
     {
         $rule = new self();
-        $rule->ruleSet = $rules;
+        $rule->ruleSet = $ruleSet;
         return $rule;
     }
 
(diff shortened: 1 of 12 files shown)


src/Rule/Each.php
```diff
@@ -17,15 +17,15 @@ final class Each extends Rule
...
     use HasValidationErrorMessage;
 
-    private Rules $rules;
+    private RuleSet $ruleSet;
 
     private string $incorrectInputMessage = 'Value should be array or iterable.';
     private string $message = '{error} {value} given.';
 
-    public static function rule(Rules $rules): self
+    public static function rule(RuleSet $rules): self
```
vjik, line 25:
```suggestion
    public static function rule(RuleSet $ruleSet): self
```

src/Rule/Each.php
```diff
@@ -17,15 +17,15 @@ final class Each extends Rule
...
     {
         $rule = new self();
-        $rule->rules = $rules;
+        $rule->ruleSet = $rules;
```
vjik, line 28:
```suggestion
        $rule->ruleSet = $ruleSet;
```

vjik (approved):

src/Rule/Each.php
arogachev: :heavy_check_mark:

src/Rule/Each.php
arogachev: :heavy_check_mark:

samdark (approved):

PR #193, opened 2022-03-16
Add support nested form.

terabytesoftw:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

### DIFF
--- a/src/Rule/Nested.php
+++ b/src/Rule/Nested.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Yiisoft\Validator\Rule;
 
+use Closure;
 use InvalidArgumentException;
 use Traversable;
 use Yiisoft\Arrays\ArrayHelper;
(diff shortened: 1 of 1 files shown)


src/Rule/Nested.php
```diff
@@ -95,8 +97,18 @@ protected function validateValue($value, ValidationContext $context = null): Res
 
             $rules = is_array($rules) ? $rules : [$rules];
             $ruleSet = new RuleSet($rules);
-            $validatedValue = ArrayHelper::getValueByPath($value, $valuePath);
+
+            if ($context !== null && $context->getDataSet() !== null) {
+                $validatedValue = $context->getDataSet()->getAttributeValue(
+                    $context->getAttribute() . '.' . $valuePath
```
samdark, line 103: That should be reflected in README, I think.

samdark:
It is a good thing that needs to be finished.

xepozz:
And please create a test with example

samdark:
Is it similar to #279?

samdark:
Yes. Closed in favor of #279.

Issue #190, opened 2022-03-12
Enhance tests


xepozz:
Test cases contain a lot of duplicated thinks.
For example let's see on https://github.com/yiisoft/validator/blob/045fd33228aa15a0f62eab5912fc3a22a027fc01/tests/Rule/EmailTest.php

1. It has multiple test cases in one method. Lines 18 and 34.
2. It has a lot of test asserts not depended between themself. From 20 to 32, from 36 to 64 lines.

Almost all test has these problems. 

The way to fix them:
1. Create new test case for each test suite.
2. Use data providers to avoid code duplication (`$validator->validate(..)->isValid()` and `$this->assert*()` multiple times)

PR #199, opened 2022-04-04
Enhance tests

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #190


samdark (approved):

tests/Rule/EmailTest.php
```diff
@@ -5,208 +5,222 @@
...
-     * Test malicious email addresses that can be used to exploit SwiftMailer vulnerability CVE-2016-10074 while IDN is
-     * enabled.
-     *
-     * @see https://legalhackers.com/advisories/SwiftMailer-Exploit-Remote-Code-Exec-CVE-2016-10074-Vuln.html
-     * @dataProvider malformedAddressesProvider
-     *
-     * @param string $value
+     * @dataProvider validateWithCheckDNSAndAllowNameProvider
      */
-    public function testMalformedAddressesIdnEnabled($value): void
```
vjik, line 194: Seems, in process refactoring skipped test mailformed addresses with IDN enabled.

vjik (changes requested):

tests/Rule/EmailTest.php
arogachev: `testMalformedAddressesIdnDisabled` and `testMalformedAddressesIdnEnabled` are actually exactly the same except `function_exists` check in "enabled" test. 

https://github.com/yiisoft/validator/blob/1ce067ebb08120b9c7540fc23cbfbb5000145726/tests/Rule/EmailTest.php#L179-L183
https://github.com/yiisoft/validator/blob/1ce067ebb08120b9c7540fc23cbfbb5000145726/tests/Rule/EmailTest.php#L196-L201

Thought that this was a logical bug and unsafe (exception will be thrown in "disabled" test in case of missing extension). I combined them into one test - if "intl" extension is available, it will run. 

So, maybe vice versa, it's lacking "disabled" check? But it's also missing in current test base. Probably we need to use something like this? 

https://github.com/yiisoft/validator/blob/1ce067ebb08120b9c7540fc23cbfbb5000145726/tests/Rule/UrlTest.php#L204-L213

Suggesting to solve it in other issue then.

vjik: Seems, here:

https://github.com/yiisoft/validator/blob/1ce067ebb08120b9c7540fc23cbfbb5000145726/tests/Rule/EmailTest.php#L181

IDN should be disabled.

arogachev: Ah, I forgot about this one. I moved it to test with default arguments, because after removing `enableIdh` it becomes just `new Email()`.

Issue #185, opened 2022-03-01
Add `Count` rule


arogachev:
Add support for array to `HasLength`.

For example:

https://github.com/yiisoft/validator/blob/e94fe51063a7f2073cac28a9acfaf5220d4e3646/tests/Rule/NestedTest.php#L226-L228

We also need to check that array contains exactly 3 elements. And it's quite common case. Currently this can be solved with custom rule or `Callback` rule, but it's a bit verbose.

Or it can be added to another rule if it goes against `HasLength` philosophy.

https://github.com/yiisoft/validator/blob/e94fe51063a7f2073cac28a9acfaf5220d4e3646/src/Rule/HasLength.php#L62-L65

xepozz:
> Or it can be added to another rule if it goes against `HasLength` philosophy.

I think it's right way. Just implement new `Count` rule. WDYT?

samdark:
`Count` sounds alright.

PR #204, opened 2022-04-08
Add Count rule

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #185

### DIFF
--- a/src/Rule/Count.php
+++ b/src/Rule/Count.php
@@ -11,6 +11,7 @@ use Yiisoft\Validator\FormatterInterface;
 use Yiisoft\Validator\Result;
 use Yiisoft\Validator\Rule;
 use Yiisoft\Validator\ValidationContext;
+
 use function count;
 
 /**
@@ -49,7 +50,7 @@ final class Count extends Rule
          * @var string user-defined error message used when the number of items is smaller than {@see $min}.
          */
         private string $tooFewItemsMessage = 'This value must contain at least {min, number} ' .
-        '{min, plural, one{items} other{items}}.',
+        '{min, plural, one{item} other{items}}.',
         /**
          * @var string user-defined error message used when the number of items is greater than {@see $max}.
          */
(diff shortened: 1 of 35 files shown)


arogachev:
If `$exactly` concept is OK, it can be later added to `HasLength` rule too.

src/Rule/Count.php
```diff
@@ -0,0 +1,129 @@
...
+use Attribute;
+use Countable;
+use InvalidArgumentException;
+use Yiisoft\Validator\FormatterInterface;
+use Yiisoft\Validator\Result;
+use Yiisoft\Validator\Rule;
+use Yiisoft\Validator\ValidationContext;
+use function count;
+
+#[Attribute(Attribute::TARGET_PROPERTY)]
```
rustamwin, line 16: I think it's better to place it after docs-block

src/Rule/Count.php
```diff
@@ -0,0 +1,129 @@
...
+namespace Yiisoft\Validator\Rule;
+
+use Attribute;
+use Countable;
+use InvalidArgumentException;
+use Yiisoft\Validator\FormatterInterface;
+use Yiisoft\Validator\Result;
+use Yiisoft\Validator\Rule;
+use Yiisoft\Validator\ValidationContext;
+use function count;
```
rustamwin, line 14:
```suggestion

use function count;
```

src/Rule/Count.php
```diff
@@ -49,11 +49,18 @@ public function __construct(
         /**
          * @var string user-defined error message used when the number of items is smaller than {@see $min}.
          */
-        private string $tooFewItemsMessage = 'This value should contain at least {min, number} {min, plural, one{items} other{items}}.',
+        private string $tooFewItemsMessage = 'This value must contain at least {min, number} ' .
+        '{min, plural, one{items} other{items}}.',
```
samdark, line 52:
```suggestion
        '{min, plural, one{item} other{items}}.',
```

samdark (approved):

rustamwin (approved):

samdark:
Thanks.

Issue #202, opened 2022-04-08
Add test for Email rule when intl extension is not available


arogachev:
Extracted from #199, see https://github.com/yiisoft/validator/pull/199#discussion_r845004137.

PR #209, opened 2022-04-12
Test "enableIdn" option with no intl extension available (Email, Url rules)

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #202

### DIFF
--- a/src/Rule/Email.php
+++ b/src/Rule/Email.php
@@ -21,6 +21,8 @@ use function strlen;
 #[Attribute(Attribute::TARGET_PROPERTY)]
 final class Email extends Rule
 {
+    use IdnDependencyTrait;
+
     public function __construct(
         /**
          * @var string the regular expression used to validate value.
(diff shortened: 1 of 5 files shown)


src/Rule/Email.php
```diff
@@ -5,13 +5,11 @@
 namespace Yiisoft\Validator\Rule;
```
arogachev, line 5: It was decided to use CI instead.

xepozz:
Could it be done with https://github.com/yiisoft/validator/blob/ea9d5279bed2f0213efed8df692ba241df23b0b3/tests/bootstrap.php#L8?

arogachev:
@xepozz Nope, it couldn't. Checked it again. This method only works once for Url rule. Already faced this problem.

arogachev:
Closing in favor of #277.

PR #216, opened 2022-04-13
Refactoring

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️

[Here](https://github.com/yiisoft/validator/pull/222) is another version without ValidatorStorage and using Container as a validator fetcher service.


xepozz:
I will continue develop a little bit different approach in a different branch 
https://github.com/yiisoft/validator/pull/222

PR #224, opened 2022-04-18
Add simple rule

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -


xepozz:
As an another way to get `SimpleRule` we can use composition:
https://gist.github.com/xepozz/91c6f501b1f2910833fc24948fa12874#file-boolean-php-L74

Issue #234, opened 2022-06-15
Add composite rule so as not to duplicate `skipOnEmpty`, `skipOnError` and `when` 


vjik:
For example, instead of:

```php
'name' => [
	new Required(when: fn() => $this->useName),
	new HasLength(min: 1, max: 50, skipOnEmpty: true, when: fn() => $this->useName),
],
```

use:

```php
'name' => [
	new CompositeRule(
		rules: [
			new Required(),
			new HasLength(min: 1, max: 50, skipOnEmpty: true),
		]
		when: fn() => $this->useName,		
	),
],
```

samdark:
It's already there: https://github.com/yiisoft/validator/blob/master/src/Rule/GroupRule.php

samdark:
https://github.com/yiisoft/validator#grouping-multiple-validation-rules

samdark:
Or do you mean something else?

vjik:
> It's already there: https://github.com/yiisoft/validator/blob/master/src/Rule/GroupRule.php

No. `GroupRule` is abstract.

> https://github.com/yiisoft/validator#grouping-multiple-validation-rules

`RuleSet` house is no more.

Need rule that implement `BeforeValidateInterface` and applying all nested rules.

PR #236, opened 2022-06-20
Add composite rule

dehbka:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | [234](https://github.com/yiisoft/validator/issues/234)

### DIFF
--- a/src/Rule/CompositeHandler.php
+++ b/src/Rule/CompositeHandler.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator\Rule;
+
+use Yiisoft\Validator\Exception\UnexpectedRuleException;
+use Yiisoft\Validator\Result;
+use Yiisoft\Validator\Rule\Trait\PreValidateTrait;
+use Yiisoft\Validator\ValidationContext;
+
+/**
+ * Can be used to group rules for validation by skipOnEmpty, skipOnError or when
+ *
+ * For example, we have same when closure:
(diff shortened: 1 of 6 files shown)


samdark (approved):

xepozz:
Have you checked rules that already exist? Here we have GroupRule that can solve your possible case.

vjik (commented):
Seems, we can not move prevalidate logic from `Validator`, and inherit `Composite` from `GroupRule`.

vjik:
> Have you checked rules that already exist? Here we have GroupRule that can solve your possible case.

Abstract `GroupRule` need for create class with nested rules in userland and then using him.

`Composite` need for direct creation with rules in constructor.

src/Rule/CompositeHandler.php
```diff
@@ -0,0 +1,53 @@
...
+    public function validate(mixed $value, object $rule, ?ValidationContext $context = null): Result
+    {
+        if (!$rule instanceof Composite) {
+            throw new UnexpectedRuleException(Composite::class, $rule);
+        }
+
+        $context->setParameter($this->parameterPreviousRulesErrored, true);
+
+        if ($this->preValidate($value, $context, $rule)) {
+            return  new Result();
```
rustamwin, line 48:
```suggestion
            return new Result();
```

rustamwin (approved):

samdark:
👍

PR #241, opened 2022-07-05
Implement self-validatable callback rule

xepozz:
Implemented self-validated, a.k.a. "simple rule".
It's an example to discuss shall we support this approach using validator.

Just copied almost all methods from `Callback` rule, its handler and tests.

You can compare the sources of simple and "not simple" rules.
[Original rule](https://github.com/yiisoft/validator/blob/a1ad9783ada37fc8504117fe61f4b697877dd7a7/src/Rule/Callback.php) and its [handler](https://github.com/yiisoft/validator/blob/a1ad9783ada37fc8504117fe61f4b697877dd7a7/src/Rule/CallbackHandler.php).
VS
[Simple rule](https://github.com/yiisoft/validator/blob/a1ad9783ada37fc8504117fe61f4b697877dd7a7/tests/Stub/CallbackSelfValidatableRule.php).

The main advantage to have it:
**Simple to write, read and support.** 

Even when the `validate` method doesn't have the check on type of the argument it looks worse.
1. Mixed responsibility. One class has either config and validation. Validation calls when the class passes to `$validator->validate()` method. Some of delayed calls. Isn't remember something? Yii2-queue has similar way. If you use Yii2-queue the job will be serialized and deserialized before it will be processed. Yii2-queue is being replacing with queue that won't have this drawback. So implementing the "simple" rule as I did we will back to already went mistakes.
2. Just try to inject some service into it. You need to implement something like service-aware service that will set needed properties with a lot magic under hood. It kills all "simple" approach.2. Just try to inject some service into it. You need to implement something like service-aware service that will set needed properties with a lot magic under hood. It kills all "simple" approach.

Any "simple" checks could be done with callback rule.

### DIFF
--- a/src/Rule/SelfValidatableRuleInterface.php
+++ b/src/Rule/SelfValidatableRuleInterface.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator\Rule;
+
+use Yiisoft\Validator\Result;
+use Yiisoft\Validator\RuleInterface;
+use Yiisoft\Validator\ValidationContext;
+
+/**
+ * Rule handler performs actual validation taking configuration parameters from a rule.
+ */
+interface SelfValidatableRuleInterface extends RuleInterface
+{
+    /**
+     * Validates the value.
+     *
+     * @param mixed $value Value to be validated.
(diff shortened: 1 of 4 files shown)


samdark:
Decided not to do it.

Issue #243, opened 2022-07-05
Got misunderstanding configuring rules over attributes


xepozz:
I've found a few problems:
1. All rules don't have REPEATABLE flag.
2. `Each`, `Nested` rules (but suddenly not `Group`, `Composite`) don't work at all without additional attributes.

We have great [example](https://github.com/yiisoft/validator/blob/1f21d2edca9373e90cfaeb231dc907c31b8256f3/tests/DataSet/AttributeDataSetTest.php#L22) with composition of attributes, but configuring an inheritance have to be different.

[How it's configuring now.](https://github.com/yiisoft/validator/blob/b7ec94dfa46762b73a45df396c5bbb7e38cfed17/tests/Data/Charts/ChartsData.php#L11)

How it should be configured:
```php
final class ChartsData
{
    #[Each(
        rules: [
            new HasMany(Chart::class),
        ],
        incorrectInputMessage: 'Custom message 1.',
        message: 'Custom message 2.'
    )]
    private array $charts;
}
```

But `HasMany` and `HasOne` aren't `Rules`, so they can't be passed into `rules`. We need to make another rule something like "as from $classname". The implementation and name are up to developer.

I'll make a pr with tests those should pass

xepozz:
@arogachev would you like to do it?

arogachev:
@xepozz Sorry, I don't understand it fully. Why we need to pass `HasMany` in rules like that?

samdark:
What's "REPEATABLE flag"?

xepozz:
> What's "REPEATABLE flag"?

https://www.php.net/manual/en/language.attributes.classes.php

xepozz:
> @xepozz Sorry, I don't understand it fully. Why we need to pass `HasMany` in rules like that?

Not HasMany, but another rule that can be "reference" or embedding for class

rustamwin:
> All rules don't have REPEATABLE flag.

Is there any case for this?

xepozz:
If array-like configuration supports multiple same validators attributes should support too.
Just for you:
```php
#[Number(min: 18, when: fn() => ...)]
#[Number(min: 21, when: fn() => ...)]
private $age;
```

Or you may want to use some composite rule and make one of its rules stronger:
```php
class CompositeRule
{
    public function getRules() {
        return [
            new Number(min: 18, when: fn() => ...),
            ...
        ];
    }
}

#[CompositeRule]
#[Number(min: 21, when: fn() => ...)]
private $age;
[...]

PR #249, opened 2022-07-16
Fix attributes collecting

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | https://github.com/yiisoft/validator/issues/243

- [X] Each rules can be used as an attribute
- [X] Even multiple times
- [X] `HasOne` and `HasMany` were replaced with `Embedded`

I've added TODO about collecting attributes in lazy mode. It may be useful because of using `Reflection` to get rules collection.


xepozz:
BTW, could someone fix psalm issue? :)

src/Attribute/Embedded.php
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    public function getRuleSet(): array
+    {
+        $classMeta = new ReflectionClass($this->referenceClassName);
+
+        return $this->collectAttributes($classMeta);
+    }
+
+    // TODO: use Generator to collect attributes
```
samdark, line 41: ?

src/DataSet/AttributeDataSet.php
```diff
@@ -36,82 +31,19 @@ public function getRules(): iterable
     {
         $classMeta = new ReflectionClass($this->baseAnnotatedObject);
 
-        return $this->handleAttributes($classMeta);
+        return $this->collectAttributes($classMeta);
     }
 
-    private function handleAttributes(ReflectionClass $classMeta): array
+    // TODO: use Generator to collect attributes
```
samdark, line 37: ?

andrew-demb:
@xepozz here it is

```diff
diff --git a/src/Attribute/Embedded.php b/src/Attribute/Embedded.php
index eb80e82..a568ad8 100644
--- a/src/Attribute/Embedded.php
+++ b/src/Attribute/Embedded.php
@@ -18,14 +18,14 @@ use Yiisoft\Validator\ValidationContext;
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 final class Embedded extends GroupRule
 {
+    /**
+     * @psalm-param Closure(mixed, ValidationContext):bool|null $when
+     */
     public function __construct(
         private string $referenceClassName,
         string $message = 'This value is not a valid.',
         bool $skipOnEmpty = false,
         bool $skipOnError = false,
-        /**
-         * @psalm-param Closure(mixed, ValidationContext):bool|null
-         */
         ?Closure $when = null,
     ) {
         parent::__construct($message, $skipOnEmpty, $skipOnError, $when);
diff --git a/src/Rule/GroupRule.php b/src/Rule/GroupRule.php
index 46dfccb..31f4ea6 100644
--- a/src/Rule/GroupRule.php
+++ b/src/Rule/GroupRule.php
@@ -27,7 +27,7 @@ abstract class GroupRule implements ParametrizedRuleInterface, BeforeValidationI
         private bool $skipOnEmpty = false,
         private bool $skipOnError = false,
         /**
-         * @psalm-param Closure(mixed, ValidationContext):bool|null
+         * @var Closure(mixed, ValidationContext):bool|null
          */
         private ?Closure $when = null,
     ) {

```

samdark (approved):

samdark:
Looks way better than current HasOne and HasMany 👍

samdark:
@xepozz seems there are issues not only with Psalm: https://github.com/yiisoft/validator/runs/7368352855?check_suite_focus=true

samdark:
This one includes https://github.com/yiisoft/validator/pull/244, right?

samdark:
Also it seems it will solve https://github.com/yiisoft/validator/pull/193, right?

xepozz:
Not sure about the last one, but the first has to be removed after the merge

src/Attribute/Embedded.php
xepozz: I've mentioned it in the PR description

xepozz:
> @xepozz seems there are issues not only with Psalm: https://github.com/yiisoft/validator/runs/7368352855?check_suite_focus=true

Seems nested attributes don't work below 8.1 php version
https://3v4l.org/Dj0lN

src/DataSet/AttributeDataSet.php
```diff
@@ -36,82 +31,19 @@ public function getRules(): iterable
...
-                        [$nestedRule],
-                        ...(($property->getAttributes(Each::class)[0] ?? null)?->getArguments() ?? [])
-                    );
-                }
-            }
-
-            $eachRuleFound = false;
-            $eachRules = [];
-            $attributes = $property->getAttributes();
+            $attributes = $property->getAttributes(RuleInterface::class, ReflectionAttribute::IS_INSTANCEOF);
```
rustamwin, line 42: Now we collect static properties as well?

src/Attribute/Embedded.php
```diff
@@ -0,0 +1,55 @@
...
+    private function collectAttributes(ReflectionClass $classMeta): array
+    {
+        $rules = [];
+        foreach ($classMeta->getProperties() as $property) {
+            $attributes = $property->getAttributes(RuleInterface::class, ReflectionAttribute::IS_INSTANCEOF);
+            foreach ($attributes as $attribute) {
+                /** @psalm-suppress UndefinedMethod */
```
rustamwin, line 48: It's better to avoid psalm-suppression or comment the reason.

src/DataSet/AttributeDataSet.php
xepozz: yep

samdark:
1. Need to mention PHP 8.0 limitations in readme.
2. Need it to be usable as code, not only as an attribute.

samdark:
PHP 8.1 specific tests should be separated.

src/Attribute/Embedded.php
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    public function getRuleSet(): array
+    {
+        $classMeta = new ReflectionClass($this->referenceClassName);
+
+        return $this->collectAttributes($classMeta);
+    }
+
+    // TODO: use Generator to collect attributes
```
samdark, line 41: Why if we return ruleset as array? We would not gain anything, I think.

samdark (approved):

src/Attribute/Embedded.php
xepozz: https://github.com/yiisoft/validator/issues/261

PR #255, opened 2022-07-16
Fix #251: Prefer rules to attributes

### LINKED ISSUE #251: Merge rules and attributes
If we have both rules and attributes, merge them with explicitly specified rules having priority:

```php
public function validate($data, iterable $rules = []): Result
{
    $data = $this->normalizeDataSet($data);

    if ($data instanceof RulesProviderInterface) {
        $rules = array_merge($data->getRules(), $rules); // <-- this

    }


```

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #251


samdark:
Test should be added after https://github.com/yiisoft/validator/pull/249 is done.

xepozz (approved):

rustamwin (approved):

src/Validator.php
```diff
@@ -35,7 +35,12 @@ public function validate($data, iterable $rules = []): Result
     {
         $data = $this->normalizeDataSet($data);
         if ($data instanceof RulesProviderInterface) {
-            $rules = $data->getRules();
+            $explicitRules = $rules;
+            $rules = (array) $data->getRules();
+
+            foreach ($explicitRules as $key => $value) {
+                $rules[$key] = $value;
```
vjik, line 42: Common rules (with int keys) shoud be added instead override.

PR #258, opened 2022-07-17
Fix #253: Replace formatter with yiisoft/translator

### LINKED ISSUE #253: Use yiisoft/translator directly
Need a PR with using yiisoft/translator directly instead of formatter interface in order to compare current solution with it.

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #253


src/Rule/AtLeastHandler.php
```diff
@@ -18,11 +18,11 @@ final class AtLeastHandler implements RuleHandlerInterface
...
     use EmptyCheckTrait;
 
-    private FormatterInterface $formatter;
+    private TranslatorInterface $translator;
 
-    public function __construct(?FormatterInterface $formatter = null)
+    public function __construct(?TranslatorInterface $translator = null)
     {
-        $this->formatter = $formatter ?? new Formatter();
+        $this->translator = $translator ?? new FallbackTranslator();
```
xepozz, line 25: Let's avoid this workaround with configuring https://github.com/yiisoft/translator-formatter-simple by default

xepozz (changes requested):

src/Rule/AtLeastHandler.php
samdark: Done in draft variant (tests fail for now, need to properly configure source).

xepozz: Looks good

xepozz:
Let unit tests check validator specific logic without right parameters substitution. 

If you want to test error messages translations you should add test suite with functional tests. 
What do you think?

config/bootstrap.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+use Psr\Container\ContainerInterface;
+use Yiisoft\Translator\CategorySource;
+use Yiisoft\Translator\MessageFormatterInterface;
+use Yiisoft\Translator\MessageReaderInterface;
+use Yiisoft\Translator\TranslatorInterface;
+
+return [
+    'validator' => static function (ContainerInterface $container) {
```
xepozz, line 12: Actually it's not necessary. 
We need configure category and add it as reference in the [collection](https://github.com/yiisoft/demo/blob/master/config/params.php#L93).

samdark:
1. Resolve conflicts.
2. Add config that adds a translator category for validation messages.
3. Add dependency on simple formatter.

samdark:
Closed in favor of https://github.com/yiisoft/validator/pull/293

PR #267, opened 2022-07-20
Fix #264: Remove GroupRule in favor of Composite

### LINKED ISSUE #264: Combine Composite and GroupRule rules into one
They have the same logic, but one of them is `abstract` and another is `final` :)
Looks like we can remove one of them and remove the class modifier.
I like `Composite` name more. I suggest to remove GroupRule.

@yiisoft/yii3 What do you think?

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #264


xepozz (approved):

src/Rule/Composite.php
```diff
@@ -18,7 +18,7 @@
  * Validates that the value is a valid json.
  */
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
-final class Composite implements SerializableRuleInterface, BeforeValidationInterface
+class Composite implements SerializableRuleInterface, BeforeValidationInterface
 {
     use BeforeValidationTrait;
     use HandlerClassNameTrait;
```
xepozz, line 24: You should delete it and use composite rule handler directly in this class as it was in group rule.

xepozz (changes requested):

rustamwin (approved):

xepozz (approved):

PR #268, opened 2022-07-20
phpDoc: Specifies the correct type of rules.

githubjeka:
To fix
image

### DIFF
--- a/src/Validator.php
+++ b/src/Validator.php
@@ -29,7 +29,7 @@ final class Validator implements ValidatorInterface
 
     /**
      * @param DataSetInterface|mixed|RulesProviderInterface $data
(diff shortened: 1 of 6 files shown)


rustamwin (approved):

src/Validator.php
```diff
@@ -135,7 +135,7 @@ private function validateInternal($value, iterable $rules, ValidationContext $co
     /**
      * @param array $rules
      *
-     * @return iterable<RuleInterface>
+     * @return iterable<RuleInterface|RuleInterface[]|\Closure|\Closure[]>
```
andrew-demb, line 138:
```suggestion
     * @return iterable<RuleInterface>
```

Normalization returns only rules

xepozz (approved):

PR #269, opened 2022-07-24
Fix #266: Make behavior expected when passing empty rules explicitly

### LINKED ISSUE #266: Allow use empty rules in `Validator::validate()`
We should change `iterable $rules = []` to `?iterable $rules = null`.

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #266

### DIFF
--- a/src/Validator.php
+++ b/src/Validator.php
@@ -29,7 +29,7 @@ final class Validator implements ValidatorInterface
 
     /**
      * @param DataSetInterface|mixed|RulesProviderInterface $data
-     * @param null|iterable<\Closure|\Closure[]|RuleInterface|RuleInterface[]> $rules
+     * @param iterable<\Closure|\Closure[]|RuleInterface|RuleInterface[]>|null $rules
      */
     public function validate(mixed $data, ?iterable $rules = null): Result
     {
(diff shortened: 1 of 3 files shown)


terabytesoftw:
`validator->validate($myObject, [])` -> no additional rules, for me its corret.
`validator->validate($myObject, null)` -> can be interpreted as if I want to override all the rules, which is not true.

vjik:
>  no additional rules

Rules that pass to `validate` don't add to object rules, but override object rules.

src/Validator.php
```diff
@@ -29,18 +29,21 @@ public function __construct(private RuleHandlerResolverInterface $ruleHandlerRes
...
     {
         $data = $this->normalizeDataSet($data);
-        if ($rules === [] && $data instanceof RulesProviderInterface) {
+        if ($rules === null && $data instanceof RulesProviderInterface) {
             $rules = $data->getRules();
         }
 
-        $context = new ValidationContext($this, $data);
         $compoundResult = new Result();
+        if (empty($rules)) {
```
vjik, line 42: On empty rules should be runned `processValidationResult()` ?

PR #270, opened 2022-07-24
Ability validate objects with attributes directly

rustamwin:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -


xepozz:
What a use-case for this?

README.md
```diff
@@ -344,6 +344,19 @@ $dataSet = new AttributeDataSet(new ChartsData(), $data);
 $errors = $validator->validate($dataSet)->getErrorMessagesIndexedByPath();
 ```
 
+If the object implements DataSetInterface just pass it directly
```
arogachev, line 347:
```suggestion
If the object implements `DataSetInterface` just pass it directly:
```

samdark:
Is it similar to #282?

src/DataSet/AttributeDataSet.php
```diff
@@ -25,7 +25,9 @@ final class AttributeDataSet implements RulesProviderInterface, DataSetInterface
     public function __construct(object $baseAnnotatedObject, array $data = [])
     {
         $this->baseAnnotatedObject = $baseAnnotatedObject;
-        $this->data = $data;
+        $this->data = ($data === [] && is_subclass_of($baseAnnotatedObject, DataSetInterface::class))
```
samdark, line 28: `instanceof` could be used.

samdark: We support attributes in non-`DataSetInterface`.

rustamwin: > instanceof could be used.

Done.

rustamwin: > We support attributes in non-DataSetInterface.

This code doesn't affect attributes support. It's just a convenience:

```php
$result = $validator->validate(new AttributeDataSet($dataSet));
```
instead of
```php
$result = $validator->validate(new AttributeDataSet($dataSet, $data));
```

rustamwin:
> Is it similar to https://github.com/yiisoft/validator/pull/282?

Similar, but not exactly.

samdark (approved):

samdark:
👍

Issue #225, opened 2022-04-22
Add detailed annotation for return value of `Rule::getOptions()`


vjik:
For example. `Boolean` rule:

```php
/**
 * @psalm-return array{
 *   trueValue: mixed,
 *   falseValue: mixed,
 *   strict: bool,
 *   message: string,
 * }
 */
public function getOptions(): array
{
	return array_merge(parent::getOptions(), [
		'trueValue' => $this->trueValue,
		'falseValue' => $this->falseValue,
		'strict' => $this->strict,
		'message' => $this->getFormattedMessage(),
	]);
}
 ```

Sergiayn:
I want to do it

arogachev:
The current intended usage is for frontend only. After discussion we decided to not add it for now.

rustamwin:
> The current intended usage is for frontend only and maybe for tests. So after discussion we decided to not add it for now.

https://github.com/yiisoft/form/blob/master/src/Field/Textarea.php#L221

arogachev:
@rustamwin Thanks. https://github.com/yiisoft/form/pull/226

PR #271, opened 2022-07-24
Add annotation for return value of Rule::getOptions()

Sergiayn:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #225

I didn't add annotation for this

```php
{
    $arrayOfRules = [];
    foreach ($this->rules as $rule) {
        if ($rule instanceof SerializableRuleInterface) {
            $arrayOfRules[] = array_merge([$rule->getName()], $rule->getOptions());
        } else {
            $arrayOfRules[] = [$rule->getName()];
        }
    }
    return $arrayOfRules;
}
```

### DIFF
--- a/src/Rule/Ip.php
+++ b/src/Rule/Ip.php
@@ -6,6 +6,7 @@ namespace Yiisoft\Validator\Rule;
 
 use Attribute;
 use Closure;
+use JetBrains\PhpStorm\ArrayShape;
 use RuntimeException;
 use Yiisoft\NetworkUtilities\IpHelper;
 use Yiisoft\Validator\SerializableRuleInterface;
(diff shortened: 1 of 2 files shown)


xepozz (changes requested):
It's useless because if you need to dump rules from a form/dto you need to use https://github.com/yiisoft/validator/blob/master/src/RulesDumper.php that will provide a lot of different information that you may use.

If you want to use `getOptions()` directly tell us the reason.

samdark:
@xepozz he's not using getOptions. This PR is about annotations.

xepozz:
I see. And why?
Nothing changed except added a need to change code twice. It's like to have both an annotation and a type for object.

arogachev:
Sorry, we decided to not add it for now. See https://github.com/yiisoft/validator/issues/225#issuecomment-1202088142 for explanation.

Issue #281, opened 2022-08-04
For `Equal` rule add boolean type of value


vjik:

samdark:
Would you please elaborate?

vjik:
I need check that user select checkbox "I agree".

In form model has property `bool $agree = false` and need rule for check `$agree` on true. 

Maybe create rule `IsTrue` - it is will be convient for this case.

arogachev:
User agreements are quite common, I am also for having this, especially if `Compare` rule was already split into separate ones. Is `true` enough? Can't remember cases when only `false` is needed.

PR #287, opened 2022-08-09
Add IsTrue rule

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #281

### DIFF
--- a/src/Rule/IsTrue.php
+++ b/src/Rule/IsTrue.php
@@ -13,10 +13,12 @@ use Yiisoft\Validator\Rule\Trait\RuleNameTrait;
 use Yiisoft\Validator\ValidationContext;
 
 /**
- * Checks if the value is a boolean value or a value corresponding to it.
+ * Checks if the value is a "true" boolean value or a value corresponding to it. Useful for user agreements etc.
+ *
+ * @see IsTrueHandler
  */
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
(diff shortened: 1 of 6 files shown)


src/Rule/IsTrue.php
```diff
@@ -0,0 +1,90 @@
...
+        /**
+         * @var Closure(mixed, ValidationContext):bool|null
+         */
+        private ?Closure $when = null,
+    ) {
+    }
+
+    /**
+     * @return mixed
+     */
```
samdark, line 48: This is already in the signature.

```suggestion
```

src/Rule/IsTrue.php
```diff
@@ -0,0 +1,90 @@
...
+    public function getTrueValue(): mixed
+    {
+        return $this->trueValue;
+    }
+
+    /**
+     * @return bool
+     */
```
samdark, line 56:
```suggestion
```

src/Rule/IsTrue.php
```diff
@@ -0,0 +1,90 @@
...
+    public function isStrict(): bool
+    {
+        return $this->strict;
+    }
+
+    /**
+     * @return string
+     */
```
samdark, line 64:
```suggestion
```

samdark (approved):

src/Rule/IsTrue.php
samdark, line 48:
```suggestion
```

samdark:
👍

Issue #132, opened 2021-05-15
Adjust "skipOnEmpty"


vjik:
1. Empty values always should be validate when rule using separately. For example, `(new Number())->validate('')->isValid()` should return `false`.

2. Validate or not validate value should be decided not at rule level, but above (maybe validator or form).

3. Check on empty can be differenet in cases. For example, `null` is empty, but empty string `''` is not empty.

samdark:
1. Yes and no. This behavior is expected to be default but I'm not sure "always" applies here.
2. Higher level could pass it to lower level.
3. Correct. What do you propose? Is it broken?

I'd keep the skipOnEmpty option at Rule level but ensure that default for individual rule is "false".

vjik:
> 1. Yes and no. This behavior is expected to be default but I'm not sure "always" applies here.

Maybe... But I think better construction `if (!empty($value) && $rule->validate($value)) ...` than `if ($rule->validate($value)->skipOnEmpty(true)) ...`. We clearly understand which values are empty.

> 2. Higher level could pass it to lower level.

If higher level could pass it to lower level, then we have unclear behavior in case when for some rule will be set `skipOnEmpty` manually.

> 3. Correct. What do you propose? Is it broken?

Remove `skipOnEmpty` from rule :)

samdark:
I'm fine with removing `skipOnEmpty` from rule but behavior itself should be kept. For example, it's needed for HTML form with:

1. First name (required, a-z)
2. Last name (required, a-z)
3. Patronymic (optional, a-z)

Patronymic here is optional. Means if it's empty string in the form submit, everything's OK but if it's "123" then we need to error.

See https://github.com/yiisoft/validator/pull/108

rustamwin:
Does current Rule/Handler mechanism solves this issue?

samdark:
In general, we have two modes:

1. HTML forms. There empty values are always passed.
2. API. There empty values are not passed.

To switch between the two:

1. Add ability to specify callback for `skipOnEmpty`. Return value will determine what empty is.
2. Values could be `true`, `false`, callback.
3. Could be specified at Validator and Rule level.
4. Create default callback classes for `SkipOnAbsent` (or `SkipOnNull`), `SkipOnEmpty`. Specify `SkipOnEmpty` as default at validator level.

arogachev:
> Specify SkipOnEmpty as default at validator level.

For validator I think it's better to specify it explicitly, because it can be also set differently for multiple rules.

PR #288, opened 2022-08-09
Adjust skip on empty

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #132


src/BeforeValidationInterface.php
```diff
@@ -11,7 +11,7 @@
  */
 interface BeforeValidationInterface
 {
-    public function shouldSkipOnEmpty(): bool;
+    public function shouldSkipOnEmpty($value): bool;
```
samdark, line 14: Could it be typed as bool|callable?

src/Rule/Trait/PreValidateTrait.php
```diff
@@ -18,7 +20,7 @@ private function preValidate(
         ValidationContext $context,
         BeforeValidationInterface $rule
     ): bool {
-        if ($rule->shouldSkipOnEmpty() && $this->isEmpty($value)) {
+        if ($rule->shouldSkipOnEmpty($value) === true) {
```
samdark, line 23:
```suggestion
        if ($rule->shouldSkipOnEmpty($value)) {
```

src/SkipOnAll.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+final class SkipOnAll
```
samdark, line 7: Was it meant to be named `DoNotSkip` or `NeverSkip`?

src/BeforeValidationInterface.php
arogachev: Nope. It's not `skipOnEmpty` parameter's value (as it was initially suggested, having different types), it's a validated value. Changed to `mixed` and improved naming to avoid confusion.

src/SkipOnAll.php
arogachev: Yes, you are right since it returns `false`. I choose `SkipNever` for better grouping. Should these classes be placed to a separate folder?

samdark: Logically yes. I think si.

arogachev: :heavy_check_mark:

README.md
```diff
@@ -142,12 +142,80 @@ new Number(asInteger: true, max: 100, skipOnError: true)
...
+```
+
+What exactly to consider to be empty is vague and can vary depending on a scope of usage.
+
+`skipOnEmpty` is more like a shortcut, `skipOnEmptyCallback` argument checks if a given value is empty:
+
+- If `skipOnEmpty` is `false`, `Yiisoft\Validator\SkipOnEmptyCallback\SkipNever` is used automatically for 
+`skipOnEmptyCallback` - every value is considered non-empty and validated without skipping (default).
+- If `skipOnEmpty` is `true`, `Yiisoft\Validator\SkipOnEmptyCallback\SkipOnEmpty` is used automatically for
+`skipOnEmptyCallback` - only non-empty values (`null`, `[]`. `''`) are validated.
```
samdark, line 160:
```suggestion
`skipOnEmptyCallback` - only non-empty values (not `null`, `[]`, or `''`) are validated.
```

README.md
```diff
@@ -142,12 +142,80 @@ new Number(asInteger: true, max: 100, skipOnError: true)
...
+- If `skipOnEmptyCallback` is set, it takes precedence to determine emptiness.
+
+While using first option is usually enough for HTML forms, the second one is more suitable for APIs.
```
samdark, line 163:
```suggestion
Using first option is usually good for HTML forms. The second one is more suitable for APIs.
```

README.md
```diff
@@ -142,12 +142,80 @@ new Number(asInteger: true, max: 100, skipOnError: true)
...
+The empty values can be also limited to `null` only:
+
+```php
+use Yiisoft\Validator\Rule\Number;
+use Yiisoft\Validator\SkipOnEmptyCallback\SkipOnNull;
+
+new Number(asInteger: true, max: 100, skipOnEmptyCallback: new SkipOnNull());
+```
+
+Note that in this case `skipOnEmpty` will be automatically set to `true`, so no need to do it manually.
```
samdark, line 174:
```suggestion
Note that in this case `skipOnEmpty` will be automatically set to `true`, so there is no need to do it manually.
```

README.md
```diff
@@ -142,12 +142,80 @@ new Number(asInteger: true, max: 100, skipOnError: true)
...
+
+For even more customization you can use own class implementing `__invoke()` magic method:
```
samdark, line 176:
```suggestion
For even more customization you can use your own class implementing `__invoke()` magic method:
```

src/SkipOnEmptyCallback/SkipNever.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator\SkipOnEmptyCallback;
+
+final class SkipNever
```
samdark, line 7: Maybe the following would be better name. What do you think?

```suggestion
final class SkipNone
```

samdark, line 7:
```suggestion
final class SkipNothing
```

samdark (approved):

src/SkipOnEmptyCallback/SkipNever.php
arogachev: Decided to choose `SkipNone`, also it's a shorter.

arogachev: Yes it's better because it's consistent with others - it highlights what is skipped and not when.

samdark:
👍

PR #296, opened 2022-08-22
Improve `Embedded` rule

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | -


README.md
```diff
@@ -468,22 +468,22 @@ Common flow is the same as you would use usual classes:
...
-        new Embedded(Point::class),
+        new Embedded(),
     ])]
     private array $points;
 }
 
 final class Point
 {
-    #[Embedded(Coordinates::class)]
+    #[Embedded]
```
xepozz, line 486: It should also work with specifying a class because the value may be a scalar value.

xepozz (changes requested):

README.md
vjik: `Embedded` rule for scalar value? Why? 
Can you describe example more detailed?

src/Rule/Trait/BeforeValidationTrait.php
```diff
@@ -19,6 +19,9 @@ public function getSkipOnEmpty(): bool
         return $this->skipOnEmpty;
     }
 
+    /**
+     * @psalm-suppress InvalidNullableReturnType NEED FIX BEFORE RELEASE!!!
```
samdark, line 23: ?

samdark (approved):

src/Rule/Trait/BeforeValidationTrait.php
vjik: Fix this psalm error required modify many files. I'll make this in separate PR.

vjik:
Closed in favor #297

Issue #292, opened 2022-08-20
Allow pass `RulesProviderInterface` as rules in method `Validator::validate()`


vjik:
It's allow create rule loaders that implemented `RulesProviderInterface`. For example, `YmlRulesProvider`.

PR #303, opened 2022-09-01
Improve variability of `$rules` argument in `Validator::validate()`

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #292


src/Rule/Nested.php
```diff
@@ -132,18 +135,13 @@ private function prepareRules(iterable|object|string|null $source): ?iterable
 
         if ($source instanceof RulesProviderInterface) {
             $rules = $source->getRules();
-            return $this->normalizeRules ? $this->normalizeRules($rules) : $rules;
```
vjik, line 135: Why we don't can make return here?

src/Rule/Nested.php
```diff
@@ -132,18 +135,13 @@ private function prepareRules(iterable|object|string|null $source): ?iterable
...
+        } elseif (!$source instanceof Traversable && !is_array($source)) {
+            $rules = (new AttributesRulesProvider($source, $this->rulesPropertyVisibility))->getRules();
```
vjik, line 139: Why we don't can make return here?

src/Validator.php
```diff
@@ -46,13 +49,27 @@ public function __construct(
...
+     * @param class-string|iterable<Closure|Closure[]|RuleInterface|RuleInterface[]>|RulesProviderInterface|null $rules
+     * @param int $rulesPropertyVisibility What visibility levels to use when reading rules from the class specified in
+     * `$rules` argument.
+     *
+     * @throws ContainerExceptionInterface
+     * @throws NotFoundExceptionInterface
      */
-    public function validate(mixed $data, ?iterable $rules = null): Result
-    {
+    public function validate(
```
samdark, line 59: It's not the same as in the interface: https://github.com/yiisoft/validator/blob/master/src/ValidatorInterface.php#L10

src/Rule/Nested.php
arogachev: To perform check and normalization for all cases and prevent duplication of corresponding code. Or can we guarantee that in other cases it's not needed? Maybe normalization can be skipped in certain scenarios, but what about checking? I suggest to leave it like that.

src/Rule/Nested.php
arogachev: https://github.com/yiisoft/validator/pull/303#discussion_r965789351

src/Validator.php
arogachev: Good catch! :+1: Then it's better to move it to a property like in `Nested` rule.

samdark (approved):

PR #316, opened 2022-09-10
[RFC] StopOnError rule

xepozz:
I would like to introduce the `Lazy` rule. 
With that rule we may stop validation of the value when the validating rule is failed.

It may be useful when we want to avoid heavy checks.

For example, the validation process may contain:
- Any external calls such as: DB calls, API calls, I/O calls
- Any heavy calculations

The better way to avoid these checks is placing ones to the end of the stack of validation:
- Check length
- Check by regex
- ... any "easy" checks ...
- Run heavy check

Example with code:
```php
final class LoginDTO
{
    #[Lazy([
        new HasLength(min: 3, max: 150),
        new Regex(pattern: '/[\w]+/i'),
        new ExistInDatabase()
    ])]
    private $username;
    // ...
    private $password;
}
```
The logic is the following: *if the length is incorrect or the username has incorrect characters not to call database check*.

We have `skip` or `when` functionality that may do the same, but there are a few differences:
1. Lazy stops any further checks
2. It may cut costs when the validation process has a lot of rules the need to skip
3. You don't need to implement `skip` functional even write ~~`skipOnEmpty => true`~~ `skipOnError => true`.

So @yiisoft/yii3 what do you think?

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,31 @@
+on:
+  pull_request:
+    paths-ignore:
+      - 'docs/**'
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - '.gitignore'
+      - '.gitattributes'
+      - 'infection.json.dist'
+      - 'psalm.xml'
+
+  push:
+    paths-ignore:
+      - 'docs/**'
+      - 'README.md'
+      - 'CHANGELOG.md'
+      - '.gitignore'
+      - '.gitattributes'
+      - 'infection.json.dist'
+      - 'psalm.xml'
+
+name: build
+
+jobs:
+  phpbench:
+    uses: yiisoft/actions/.github/workflows/phpbench.yml@master
+    with:
+      os: >-
+        ['ubuntu-latest', 'windows-latest']
+      php: >-
+        ['8.0', '8.1']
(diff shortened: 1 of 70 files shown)


arogachev:
1. Can't the same be achieved with `skipOnError: true`?
2. Maybe it would be even cooler to have some "weight" property for rule so the whole list can be ordered by it. With option to do it dynamically the order can be different and doesn't require the participation of an end user.

xepozz:
> * Can't the same be achieved with `skipOnError: true`?
> * Maybe it would be even cooler to have some "weight" property for each rule so the whole list can be ordered by it. With option to do it dynamically the order can be different and doesn't require the participation of an end user.

1. I wrote about it
2. It's too complicated to understand the order in common and it doesn't solve the problem

samdark:
1. The name could be better as `StopOnError` or `SkipOnError`.
2. Would it allow removing `skipOnError` from rules altogether?

xepozz:
> * The name could be better as `StopOnError` or `SkipOnError`.
> * Would it allow removing `skipOnError` from rules altogether?

1. `StopOnError` suits more.
2. Rules with `skipOnError` just skips itself when this rule stops propagation at all.

samdark (approved):
Looks good to me but let's add it README.

rustamwin (approved):

samdark (approved):

samdark:
👍

Issue #186, opened 2022-03-01
Add a more convenient way to configure the same properties for multiple rules


arogachev:
For example, formatter. For usages like this it can quickly become a hassle.

https://github.com/yiisoft/validator/blob/e94fe51063a7f2073cac28a9acfaf5220d4e3646/tests/Rule/NestedTest.php#L209-L233

arogachev:
For rule set it's possible to use common formatter though, but other properties, such as `skipOnError`, etc., still need to be duplicated.

https://github.com/yiisoft/validator/blob/e94fe51063a7f2073cac28a9acfaf5220d4e3646/src/RuleSet.php#L40

arogachev:
The same applies to `RulesDumper`:

https://github.com/yiisoft/validator/blob/e94fe51063a7f2073cac28a9acfaf5220d4e3646/src/RulesDumper.php#L72

I think the better way is to set formatter during configuration / initialization process for each rule individually.

arogachev:
`Validator` can't fully handle this too for more complex structures (`Nested` + `Each` combinations).

https://github.com/yiisoft/validator/blob/e94fe51063a7f2073cac28a9acfaf5220d4e3646/src/Validator.php#L42

samdark:
Can't you get an instance of a rule and then re-use it multiple times?

arogachev:
Was it fixed by #234? @vjik

arogachev:
Working on it.

PR #317, opened 2022-09-12
Add options' progagation for `Nested` rule

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #186

### DIFF
--- a/src/PropagateOptionsInterface.php
+++ b/src/PropagateOptionsInterface.php
@@ -1,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+interface PropagateOptionsInterface
+{
+    public function propagateOptions(): void;
+}
--- a/src/PropagateOptionsInterface.php
+++ b/src/PropagateOptionsInterface.php
@@ -4,6 +4,10 @@ declare(strict_types=1);
 
 namespace Yiisoft\Validator;
 
+/**
+ * An interface implemented by rules that can propagate their common options (such as `skipOnEmpty`, `skipOnError`,
+ * `when`) to child rules as an alternative way of specifying them explicitly in every child rule.
+ */
 interface PropagateOptionsInterface
 {
     public function propagateOptions(): void;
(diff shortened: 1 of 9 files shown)


src/PropagateOptionsInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator;
+
+interface PropagateOptionsInterface
```
samdark, line 7: Need phpdoc.

src/Rule/Each.php
```diff
@@ -20,15 +21,19 @@
...
+    PropagateOptionsInterface
 {
     use BeforeValidationTrait;
     use RuleNameTrait;
     use SkipOnEmptyTrait;
 
     public function __construct(
         /**
-         * @var iterable<RuleInterface>
+         * @var iterable<BeforeValidationInterface|RuleInterface|SkipOnEmptyInterface>
```
samdark, line 36: Does the rule always implement all these interfaces?

arogachev: Yes. It was added specifically for Psalm. Maybe there is a simpler way?

src/Rule/Nested.php
```diff
@@ -80,6 +85,7 @@ public function __construct(
         private bool $requirePropertyPath = false,
         private string $noPropertyPathMessage = 'Property path "{path}" is not found.',
         private bool $normalizeRules = true,
+        private bool $propagateOptions = false,
```
samdark, line 88: Is there a reason it's `false` by default? Are there valid cases when we want it `false`?

arogachev: To configure other `Nested` rules within base `Nested` rule separately like these ones:

- https://github.com/yiisoft/validator/blob/5d1a1f48422cf44bd050564086ad79a7aa167695/tests/Rule/NestedHandlerTest.php#L505
- https://github.com/yiisoft/validator/blob/5d1a1f48422cf44bd050564086ad79a7aa167695/tests/Rule/NestedHandlerTest.php#L508

Also otherwise in the same example options [here](https://github.com/yiisoft/validator/blob/5d1a1f48422cf44bd050564086ad79a7aa167695/tests/Rule/NestedHandlerTest.php#L508) will be set 3 times.

samdark (approved):

PR #319, opened 2022-09-14
Move translator to `ValidationContext`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | -

### DIFF
--- a/src/SimpleRuleHandlerContainer.php
+++ b/src/SimpleRuleHandlerContainer.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Yiisoft\Validator;
 
-use Error;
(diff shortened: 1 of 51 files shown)


src/SimpleRuleHandlerContainer.php
```diff
@@ -30,7 +25,7 @@ public function resolve(string $className): RuleHandlerInterface
         }
 
         try {
-            $classInstance = new $className(translator: $this->translator);
+            $classInstance = new $className();
```
arogachev, line 28: I guess `try` / `catch` could be removed completely with new approach?

samdark:
https://github.com/yiisoft/validator/pull/320 looks like a better option.

PR #320, opened 2022-09-14
[RFC] Split validator features

xepozz:
- [X] Translator is a optional dependency now
- [X] Translations moved out the validator into separated class decorator
- [X] Added ErrorMessage class that allows to store message template and parameters independently. Think it may be useful to use both not touched messages, formatted with `simple formatter` and translated.
- [X] Refactored tests
- Using the validator without installation the `translator` packages gives not formatted error messages even without any parameters substitutions, so user will get `Value {value} is bad` message instead of `Value 5 is bad`. Parameters also available to do substitution.


samdark:
Overall I like it.

# Advantages

1. Simpler rule handlers.
2. Clean rule handler constructor.
3. Simpler code overall.

# Disadvantages

1. It's impossible to extract messages now by analyzing `->translate()` calls. That's the problem of extractor tool so we may come up with a way to tag strings as translation IDs.
2. Likely unusual configuration (but overall it's alright).
3. Ability not to use one of the decorators. That's not great especially in case of normalizer.
4. Tests are less readable.

# Things to consider

1. Naming. `ValidatorDecorator` posifix looks like an overkill.
2. Default config should be adjusted.

xepozz:
> It's impossible to extract messages now by analyzing ->translate() calls. That's the problem of extractor tool so we may come up with a way to tag strings as translation IDs.

Actually it's possible with an additional decorator that will capture strings and provide them to the extractor

rustamwin:
What is the purpose/benefit of adding `ErrorMessage`?

src/Result.php
```diff
@@ -45,29 +45,29 @@ public function getErrors(): array
     }
 
     /**
-     * @return string[]
+     * @return ErrorMessage[]
```
samdark, line 48: That makes using results a bit more complicated.

src/DatasetNormalizerValidatorDecorator.php
```diff
@@ -0,0 +1,80 @@
...
+use Yiisoft\Validator\DataSet\MixedDataSet;
+use Yiisoft\Validator\DataSet\ObjectDataSet;
+use Yiisoft\Validator\RulesProvider\AttributesRulesProvider;
+
+use function is_array;
+
+/**
+ * Validator validates {@link DataSetInterface} against rules set for data set attributes.
+ */
+final class DatasetNormalizerValidatorDecorator implements ValidatorInterface
```
samdark, line 23: I still think it doesn't make sense to decouple it from the validator itself since without normalizer validator isn't really usable.

tests/DataSet/PHP80/ObjectDataSet80Test.php
```diff
@@ -144,7 +145,10 @@ public function testValidateWithCallbackMethod(): void
         $this->assertInstanceOf(Callback::class, $rules['name'][0]);
 
         $result = $validator->validate(['name' => 'bar'], $rules);
-        $this->assertSame(['name' => ['Value must be "foo"!']], $result->getErrorMessagesIndexedByPath());
+        $this->assertSame(['name' => ['Value must be "foo"!']], array_map(
```
samdark, line 148: This is quite bad UX, I think.

tests/Rule/EachHandlerTest.php
```diff
@@ -34,7 +39,10 @@ public function testErrorMessagesIndexedByPath(object $rule, $value, array $expe
         $result = $this->validate($value, $rule);
 
         $this->assertFalse($result->isValid(), print_r($result->getErrorMessagesIndexedByPath(), true));
-        $this->assertEquals($expectedErrors, $result->getErrorMessagesIndexedByPath());
+        $this->assertEquals($expectedErrors, array_map(
+            fn (array $errors) => array_map(fn (ErrorMessage $error) => $error->getMessage(), $errors),
+            $result->getErrorMessagesIndexedByPath()
+        ));
```
samdark, line 45: This is quite bad UX, I think.

tests/Rule/NestedTest.php
```diff
@@ -237,19 +238,19 @@ public function dataNestedWithoutRulesToNonObject(): array
...
+//    public function testNestedWithoutRulesToNonObject(string $expectedValueName, object $data): void
+//    {
+//        $validator = $this->createValidator();
+//
+//        $this->expectException(InvalidArgumentException::class);
+//        $this->expectExceptionMessage(
+//            'Nested rule without rules could be used for objects only. ' . $expectedValueName . ' given.'
+//        );
+//        $validator->validate($data);
+//    }
```
samdark, line 253: ?

tests/Rule/NestedTest.php
```diff
@@ -403,11 +404,14 @@ public function testHandler(
 
         $this->assertSame($expectedIsValid, $result->isValid());
         if (!$expectedIsValid) {
-            $this->assertSame($expectedErrorMessagesIndexedByPath, $result->getErrorMessagesIndexedByPath());
+            $this->assertEquals($expectedErrorMessagesIndexedByPath, array_map(
+                fn (array $errors) => array_map(fn (ErrorMessage $error) => $error->getMessage(), $errors),
+                $result->getErrorMessagesIndexedByPath()
+            ));
```
samdark, line 410: This is quite bad UX, I think.

tests/ValidateLogicTest.php
```diff
@@ -147,7 +148,13 @@ public function getRules(): iterable
...
     {
         $result = $this->createValidator()->validate($data, $rules);
-        $this->assertSame($expectedErrorMessages, $result->getErrorMessagesIndexedByAttribute());
+        $this->assertEquals(
+            $expectedErrorMessages,
+            array_map(
+                fn (array $errors) => array_map(fn (ErrorMessage $error) => $error->getMessage(), $errors),
+                $result->getErrorMessagesIndexedByAttribute(),
+            )
+        );
```
samdark, line 157: This is quite bad UX, I think.

src/Result.php
xepozz: It's almost the same thanks to `__toString()` method inside `ErrorMessage` and thanks to it `TranslatorDecorator` can translate error messages

tests/ValidateLogicTest.php
xepozz: It returns all parts of `ErrorMessage` to satisfy user's needs if the one wants to expose raw value to a frontend.
Something like that. If you have better design decision describe it please.

src/ErrorMessage.php
```diff
@@ -0,0 +1,35 @@
...
+    {
+        return $this->message;
+    }
+
+    public function getParameters(): array
+    {
+        return $this->parameters;
+    }
+
+    public function __toString(): string
```
samdark, line 31: It doesn't do replacement of parameters into message so not the same as ready to use messages.

src/Validator.php
```diff
@@ -62,19 +61,22 @@ public function __construct(
...
         } elseif ($rules instanceof RulesProviderInterface) {
             $rules = $rules->getRules();
         } elseif (!$rules instanceof Traversable && !is_array($rules) && $rules !== null) {
             $rules = (new AttributesRulesProvider($rules, $this->rulesPropertyVisibility))->getRules();
         }
+        $context = new ValidationContext($context?->getValidator() ?? $this, $context?->getDataSet() ?? null, null, $context?->getParameters() ?? []);
 
         $compoundResult = new Result();
-        $context = new ValidationContext($this, $data);
+        $context = new ValidationContext($context?->getValidator() ?? $this, $context?->getDataSet() ?? $data, $context?->getAttribute() ?? null, $context?->getParameters() ?? []);
```
samdark, line 79:
```suggestion
        $context = new ValidationContext($context?->getValidator() ?? $this, $context?->getDataSet() ?? $data, $context->getAttribute() ?? null, $context?->getParameters() ?? []);
```

src/TranslateValidatorDecorator.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function validate(
+        mixed $data,
+        iterable|RulesProviderInterface|null $rules = null,
+        ?ValidationContext $context = null,
+    ): Result {
+        $context = new ValidationContext($context?->getValidator() ?? $this, $context?->getDataSet() ?? null, null, $context?->getParameters() ?? []);
+        $result = $this->decorated->validate($data, $rules, $context);
+
+        $isTranslationNeeded = $context->getParameter(self::IS_TRANSLATION_NEEDED, true) ?? true;
```
rustamwin, line 40:
```suggestion
        $isTranslationNeeded = $context->getParameter(self::IS_TRANSLATION_NEEDED, true);
```

src/ErrorMessage.php
xepozz: It won't do. Message formatting is a part of `translator` especially `MessageFormatterInterface`.

So if you want either formatted or translated messages you need to add translator as a dependency, configure it and add translator decorator to validator

xepozz: The reason is ability to expose raw error messages with their's parameters. If it can't be, frontend or third-party libraries will get already formatted (with simple formatter or something else) message. For example, `Value 55 is bad` instead of `Value {value} is bad`.
Otherwise it can be solved with saving the original error message that is worse than the current design.

composer.json
```diff
@@ -35,11 +34,13 @@
...
         "vimeo/psalm": "^4.18",
-        "yiisoft/test-support": "^3.0"
+        "yiisoft/test-support": "^3.0",
+        "yiisoft/translator": "^1.1"
     },
     "suggest": {
         "ext-intl": "Allows using IDN validation for emails",
-        "yiisoft/di": "To create rule handlers via Yii DI"
+        "yiisoft/di": "To create rule handlers via Yii DI",
+        "yiisoft/translator": "Formats and translates validator messages"
```
rustamwin, line 43:
```suggestion
        "yiisoft/translator": "To format and translate validator messages"
```

rustamwin (approved):

Issue #289, opened 2022-08-15
Differentiate when value is not passed / `null` is passed


arogachev:
Fixing #173 is fine but even without raising exception it could be useful.

PR #322, opened 2022-09-21
Differentiate when value is not passed / `null` is passed

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #289


src/Rule/Trait/PreValidateTrait.php
```diff
@@ -23,8 +21,8 @@ private function preValidate(
         BeforeValidationInterface $rule
     ): bool {
         if (
-            $rule instanceof SkipOnEmptyInterface
-            && (SkipOnEmptyNormalizer::normalize($rule->getSkipOnEmpty()))($value)
+            $rule instanceof SkipOnEmptyInterface &&
+            (SkipOnEmptyNormalizer::normalize($rule->getSkipOnEmpty()))($value, $rule, $context)
```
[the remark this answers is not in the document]
arogachev: Sorry, what do you mean?

src/DataSet/ObjectDataSet.php
```diff
@@ -66,6 +66,11 @@ public function getAttributeValue(string $attribute): mixed
             : null;
     }
 
+    public function hasAttribute(string $attribute): bool
+    {
+        return true;
```
xepozz, line 71: I think we also need to check the value if it has the attribute as you did it ArrayDataSet.

xepozz (changes requested):

src/Rule/Trait/PreValidateTrait.php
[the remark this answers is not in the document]
arogachev: Sorry, still have not understood. The only changes were moving logical operator (code style) and adding missing arguments. Maybe you can illustrate with example?

src/DataSet/ObjectDataSet.php
arogachev: Agree. Also found another bug during making these changes. Will commit fix soon.

arogachev: Fixed.

src/DataSet/ArrayDataSet.php
```diff
@@ -14,11 +16,16 @@ public function __construct(private array $data = [])
 
     public function getAttributeValue(string $attribute): mixed
     {
-        return $this->data[$attribute] ?? null;
+        return $this->hasAttribute($attribute) ? $this->data[$attribute] : null;
```
vjik, line 19:
```suggestion
        return $this->data[$attribute] ?? null;
```

Previous code is more elegant.

src/DataSet/ObjectDataSet.php
```diff
@@ -77,9 +86,6 @@ private function parseObject(): void
         $this->rules = $objectHasRules ? $this->object->getRules() : [];
 
         $this->dataSetProvided = $this->object instanceof DataSetInterface;
-        if ($this->dataSetProvided) {
-            return;
-        }
```
vjik, line 82: If object instance of `DataSetInterface` then we don't need parse object.

src/Rule/RequiredHandler.php
```diff
@@ -31,15 +29,24 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
...
+        }
 
-        if ($this->isEmpty(is_string($value) ? trim($value) : $value)) {
-            $formattedMessage = $this->translator->translate(
-                $rule->getMessage(),
-                ['attribute' => $context->getAttribute(), 'value' => $value]
-            );
-            $result->addError($formattedMessage);
+        if (is_string($value)) {
+            $value = trim($value);
```
vjik, line 40: Maybe make `trim` optional? 
Suggest make this in another PR.

src/DataSet/ObjectDataSet.php
arogachev: No, we do need to parse the object at least for the rules. Otherwise some tests will fail.

Empty rules, no errors while rules are declared via attributes:

https://github.com/yiisoft/validator/blob/616987bfef8820a5292086064308b918af88c51d/tests/ValidatorTest.php#L291-L299
https://github.com/yiisoft/validator/blob/616987bfef8820a5292086064308b918af88c51d/tests/DataSet/ObjectDataSetTest.php#L79

The first test is new, and the second is existing but assert was wrong.

src/Rule/RequiredHandler.php
arogachev: `trim` was already there before these changes. Can be seen as new because of the refactoring. Yes, suggest to improve it in a separate PR. I actually had it written in my notes, created the issue - #328.

src/DataSet/MixedDataSet.php
```diff
@@ -21,4 +24,9 @@ public function getData(): mixed
     {
         return $this->value;
     }
+
+    public function hasAttribute(string $attribute): bool
+    {
+        return true;
```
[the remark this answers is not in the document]
arogachev: Agree, it was even added in the description by me. :slightly_smiling_face:  As far as I remember, `false` was breaking tests, so I thought it was more a naming problem. But I checked the tests again and it's alright. Thanks for noticing. :+1:

src/Rule/Trait/PreValidateTrait.php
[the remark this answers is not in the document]
arogachev: This was one of the options I initially considered. But then I thought they could be useful for some more customization. And the signature also matches custom callable signature for `Callback` rule. But maybe they are indeed redundant until a practical use case is found. Let's only leave `$isAttributeMissing` for now.

src/DataSet/ObjectDataSet.php
vjik: If object provide data via implementation of `DataSetInterface` then we don't can parse rules from properties, because data is not properties.

src/Rule/Trait/PreValidateTrait.php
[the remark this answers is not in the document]
arogachev: Need to adjust in other places within a whole project.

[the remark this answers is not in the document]
arogachev: Done.

src/DataSet/ArrayDataSet.php
arogachev: Yes, using `hasAttribute()` outside of this scope is enough. Adjusted in other places as well.

Issue #302, opened 2022-09-01
Cache rules in `ObjectDataSet`


arogachev:
- https://t.me/yii3ru/74674.

arogachev:
Suggested by @BoShurik.

PR #323, opened 2022-09-27
Add caching to ObjectDataSet

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #302


BoShurik:
Is it possible to use PSR Cache interface so we can persist state between requests?
E.g.
```
ObjectDataSet::setCache(CacheInterface $cache);
```

arogachev:
Actually I was thinking about it. But what about serialization (`Callback` rules for example)?

BoShurik:
Ignore caching in that case

arogachev:
Thanks, we'll consider this. What do you guys think?

BoShurik:
> Ignore caching in that case

Fallback to `ArrayAdapter`

xepozz:
Could you move caching into separated class and use either Decorator or Composition patterns?

samdark:
See #320 about @xepozz idea.

samdark:
I think it won't give any profit storing it in Redis / DB because of unserialization cost. I'd keep it as in-memory cache for either a single request-response (in classic fpm mode) or long running context (for Swoole/RR).

samdark (approved):

xepozz:
> I think it won't give any profit storing it in Redis / DB because of unserialization cost. I'd keep it as in-memory cache for either a single request-response (in classic fpm mode) or long running context (for Swoole/RR).

Agree that it makes sense only with memoization, but it still should be done as a separated class over original `ObjectDataSet`.

rustamwin (approved):

src/DataSet/ObjectDataSet.php
```diff
@@ -61,41 +67,70 @@ public function getAttributeValue(string $attribute): mixed
...
+        if ($this->dataSetProvided) {
+            $data = $this->object->getData();
+        } else {
+            $data = [];
+            foreach ($this->reflectionProperties as $name => $property) {
+                $data[$name] = $property->getValue($this->object);
+            }
         }
+
+        $this->updateCacheItem('data', $data);
```
vjik, line 89: Seems here may be collision when use several instances of `ObjectDataSet` for same object class. Maybe save `$data` to non-static private variable?

arogachev:
OK I'll add a decorator.

src/DataSet/ObjectDataSet.php
rustamwin: What about using `spl_object_id` for the cache key.

src/DataSet/CacheObjectDataSetDecorator.php
```diff
@@ -0,0 +1,91 @@
+<?php
```
arogachev, line 1: @vjik's suggestion: Do not use a decorator.

src/ObjectDataSetInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
```
arogachev, line 1: @vjik's suggestion: Do not use an additional interface.

src/DataSet/ObjectDataSet.php
```diff
@@ -9,43 +9,114 @@
...
-    public function getObject(): object
+    public function getRules(): iterable
     {
-        return $this->object;
+        if ($this->rules !== null) {
+            return $this->rules;
+        }
+
+        $objectHasRules = $this->object instanceof RulesProviderInterface;
+        $rules = $objectHasRules ? $this->object->getRules() : [];
```
arogachev, line 50: @vjik's suggestion: Do not perform caching via static variable when object has its own `getRules()` implementation.

xepozz:
if we do memoization than it's enough to use decorator that will cache only `getRules()` call of decorated object.

src/DataSet/ObjectDataSet.php
```diff
@@ -54,64 +129,52 @@ public function getAttributeValue(string $attribute): mixed
...
             ? $this->object->hasAttribute($attribute)
-            : array_key_exists($attribute, $this->reflectionProperties);
+            : array_key_exists($attribute, $this->getReflectionProperties());
     }
 
     public function getData(): array
     {
-        if ($this->dataSetProvided) {
-            return $this->object->getData();
+        if ($this->data !== null) {
```
vjik, line 144: If object provided `DataSetInterface` then not need cache and need use `getData()` from object.

```php
if ($this->dataSetProvided) {
        return $this->object->getData();
}
````

File: src/DataSet/ObjectDataSet.php [hunk 6]
```diff
@@ -13,39 +14,113 @@
...
     public function __construct(
         private object $object,
-        private int $propertyVisibility = ReflectionProperty::IS_PRIVATE|ReflectionProperty::IS_PROTECTED|ReflectionProperty::IS_PUBLIC
+        private int $propertyVisibility = ReflectionProperty::IS_PRIVATE |
+        ReflectionProperty::IS_PROTECTED |
+        ReflectionProperty::IS_PUBLIC
     ) {
-        $this->parseObject();
+        $this->dataSetProvided = $this->object instanceof DataSetInterface;
+        $this->cacheKey = get_class($this->object);
```
vjik, line 50: Cache key must take into account class name and property visibility.

vjik (changes requested):

src/DataSet/ObjectDataSet.php
arogachev: OK, but unfortunately `benchGetRulesWithNewInstanceAndNoCache()` no longer works correctly. Can be removed for now I guess, since profit of cache is already proven.

src/DataSet/ObjectDataSet.php
```diff
@@ -54,16 +123,14 @@ public function getAttributeValue(string $attribute): mixed
             return $this->object->getAttributeValue($attribute);
         }
 
-        return $this->hasAttribute($attribute)
-            ? $this->reflectionProperties[$attribute]->getValue($this->object)
-            : null;
+        return $this->getData()[$attribute] ?? null;
```
vjik, line 126:
```suggestion
        return ($this->getReflectionProperties()[$attribute] ?? null)?->getValue($this->getObject());
```

vjik: Not need get all data when receiving one attribute value.

src/DataSet/ObjectDataSet.php
```diff
@@ -16,36 +18,103 @@
...
+
+        if ($this->canCache()) {
+            $this->setCacheItem('rules', $rules);
+        }
+
+        return $rules;
     }
 
-    public function getRules(): iterable
+    private function getReflectionProperties(): array
```
vjik, line 91: Private method should be placed after public methods.

tests/DataSet/ObjectDataSetTest.php
```diff
@@ -268,6 +332,70 @@ public function testValidateWithWrongCallbackMethod(): void
                 $object::class,
             )
         );
-        new ObjectDataSet($object);
+        (new ObjectDataSet($object))->getRules();
```
vjik, line 335: We testing `getRules()`. `ObjectDataSet` should be created before asserts.

arogachev: It's already fixed. Maybe diff was shown incorrectly.

src/DataSet/ObjectDataSet.php
arogachev: Good catch. :+1: It was left since time when the data was cached.

vjik (approved):

PR #324, opened 2022-09-27
Cleanup and minor refactoring

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |


tests/Stub/CustomUrlRule.php
```diff
@@ -1,25 +0,0 @@
...
-declare(strict_types=1);
-
-namespace Yiisoft\Validator\Tests\Stub;
-
-use Yiisoft\Validator\Rule\Composite;
-use Yiisoft\Validator\Rule\HasLength;
-use Yiisoft\Validator\Rule\Required;
-use Yiisoft\Validator\Rule\Url;
-
-final class CustomUrlRule extends Composite
```
xepozz, line 12: I remember it was in use but I've checked right now and it's not. Better to restore tests with this rule.

samdark (approved):

tests/Stub/CustomUrlRule.php
arogachev: OK. Suggest to revert the changes and restore the tests in a separate PR.

Issue #329, opened 2022-10-06
Split `BeforeValidationInterface` for separate interface by feature


vjik:
Example of separate interface: [SkipOnEmptyInterface](https://github.com/yiisoft/validator/blob/master/src/SkipOnEmptyInterface.php).

`BeforeValidationInterface` should be split to:
- `SkipOnErrorInterface`
- `WhenInterface`

Issue #326, opened 2022-10-05
In `Nested` and `Each` rule propagate can be error when rule don't implement `BeforeValidationInterface`


vjik:
https://github.com/yiisoft/validator/blob/aa582666544411b0b22eee7745650c44f83b332d/src/Rule/Each.php#L58-L60

https://github.com/yiisoft/validator/blob/aa582666544411b0b22eee7745650c44f83b332d/src/Rule/Nested.php#L233-L235

PR #331, opened 2022-10-07
Split `BeforeValidationInterface` to separate interfaces

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ✔️
| Fixed issues  | #326 #329

### DIFF
--- a/src/Rule/Nested.php
+++ b/src/Rule/Nested.php
@@ -11,18 +11,20 @@ use JetBrains\PhpStorm\ArrayShape;
 use ReflectionProperty;
 use Traversable;
 use Yiisoft\Strings\StringHelper;
-use Yiisoft\Validator\BeforeValidationInterface;
 use Yiisoft\Validator\PropagateOptionsInterface;
-use Yiisoft\Validator\Rule\Trait\BeforeValidationTrait;
 use Yiisoft\Validator\Rule\Trait\RuleNameTrait;
 use Yiisoft\Validator\Rule\Trait\SkipOnEmptyTrait;
+use Yiisoft\Validator\Rule\Trait\SkipOnErrorTrait;
+use Yiisoft\Validator\Rule\Trait\WhenTrait;
 use Yiisoft\Validator\RuleInterface;
 use Yiisoft\Validator\RulesDumper;
 use Yiisoft\Validator\RulesProvider\AttributesRulesProvider;
 use Yiisoft\Validator\RulesProviderInterface;
 use Yiisoft\Validator\SerializableRuleInterface;
 use Yiisoft\Validator\SkipOnEmptyInterface;
+use Yiisoft\Validator\SkipOnErrorInterface;
 use Yiisoft\Validator\ValidationContext;
+use Yiisoft\Validator\WhenInterface;
 
 use function array_pop;
 use function count;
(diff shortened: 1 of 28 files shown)


rustamwin (approved):

src/Rule/Nested.php
```diff
@@ -167,7 +171,11 @@ private static function ensureArrayHasRules(iterable &$rules)
                 continue;
             }
             if (!$rule instanceof RuleInterface) {
-                $message = sprintf('Each rule should be an instance of %s, %s given.', RuleInterface::class, get_debug_type($rule));
```
arogachev, line 170: Ideally this should be handled by a linter - provide a warning about exceeding max line length or split arguments automatically.

src/Rule/Trait/PreValidateTrait.php
```diff
@@ -27,10 +27,18 @@ private function preValidate(
...
+            $rule instanceof SkipOnErrorInterface
+            && $rule->shouldSkipOnError()
+            && $context->getParameter($this->parameterPreviousRulesErrored) === true
+        ) {
             return true;
         }
 
-        return is_callable($rule->getWhen()) && !($rule->getWhen())($value, $context);
+        if ($rule instanceof WhenInterface) {
+            return ($rule->getWhen() !== null) && !($rule->getWhen())($value, $context);
```
arogachev, line 39:
```suggestion
            $when = $rule->getWhen();

            return $when !== null && !$when($value, $context);
```

arogachev: OR at least the first pair of parentheses can be omitted.

tests/ValidatorTest.php
```diff
@@ -171,9 +169,7 @@ public function requiredDataProvider(): array
             'sort' => [
                 new InRange(
                     ['asc', 'desc'],
-                    skipOnEmpty: static function (mixed $value, bool $isAttributeMissing): bool {
-                        return $isAttributeMissing || $value === '';
-                    }
+                    skipOnEmpty: static fn (mixed $value, bool $isAttributeMissing): bool => $isAttributeMissing || $value === ''
```
arogachev, line 172: This one is too long now. We actually decided to disable this rule completely because it also transforms multiple line static functions like that. Ideally need to configure it to transform only when the result fits max line length.

Can be disabled like that - https://github.com/yiisoft/yii-runner-console/pull/22/commits/ff8151d696cad3464974fa6da99e52fb78acfc11.

arogachev (changes requested):
:+1:

src/Rule/Nested.php
arogachev: https://github.com/yiisoft/docs/issues/175

src/Rule/Trait/PreValidateTrait.php
vjik: Done

tests/ValidatorTest.php
vjik: Fix line lenght in this case.

tests/ValidatorTest.php
```diff
@@ -169,7 +169,10 @@ public function requiredDataProvider(): array
             'sort' => [
                 new InRange(
                     ['asc', 'desc'],
-                    skipOnEmpty: static fn (mixed $value, bool $isAttributeMissing): bool => $isAttributeMissing || $value === ''
+                    skipOnEmpty: static fn (
+                        mixed $value,
+                        bool $isAttributeMissing
+                    ): bool => $isAttributeMissing || $value === ''
```
arogachev, line 175:
```suggestion
                    skipOnEmpty: static function (mixed $value, bool $isAttributeMissing): bool {
                        return $isAttributeMissing || $value === '';
                    }
```

arogachev: (Optional, code style) The style before the changes is better for one-liners for my opinion.

arogachev (approved):

PR #332, opened 2022-10-07
Fix tests

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | 

I've removed any `function_exists` or `extension_loaded` conditions and added my own tool to mock internal functions.
With that we can remove conditional run of tests in CI and all test can be passed in local machine too.


src/Rule/Email.php
```diff
@@ -15,8 +15,6 @@
 use Yiisoft\Validator\SkipOnEmptyInterface;
 use Yiisoft\Validator\ValidationContext;
 
-use function function_exists;
-
```
vjik, line 19: Can we make this without remove `use`?

xepozz: You can't mock internal function if it already exists

vjik (changes requested):
Not using qualified functions just because of the tests is a bad solution.

xepozz:
Hmm, looks like PHPUnit on Windows cannot load bootstrap file at startup.

arogachev:
CI approach may be not ideal in terms of best practices, but has proved itself as a working solution with minimum source / tests code modifications. Closing for now.

PR #341, opened 2022-10-14
`ObjectDataSet` cache suggestion

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/src/DataSet/ObjectDataSet.php
+++ b/src/DataSet/ObjectDataSet.php
@@ -13,106 +13,48 @@ use Yiisoft\Validator\RuleInterface;
 use Yiisoft\Validator\RulesProviderInterface;
 
 use function array_key_exists;
-use function get_class;
 
 /**
  * This data set makes use of attributes introduced in PHP 8. It simplifies rules configuration process, especially for
- * nested data and relations. Please refer to the guide for examples.
+ * nested data and relations. Please refer to the guide for example.
  *
  * @link https://www.php.net/manual/en/language.attributes.overview.php
  */
 final class ObjectDataSet implements RulesProviderInterface, DataSetInterface
 {
     private bool $dataSetProvided;
-    /**
-     * @var ReflectionProperty[]
-     */
-    private ?array $reflectionProperties = null;
-    private ?array $data = null;
+    private bool $rulesProvided;
 
-    private static array $cachedRules = [];
-    private string $cacheKey;
+    private ?array $cachedData = null;
+
+    private static array $cache = [];
+    private ?string $cacheKey;
 
     public function __construct(
         private object $object,
-        private int $propertyVisibility = ReflectionProperty::IS_PRIVATE |
-        ReflectionProperty::IS_PROTECTED |
-        ReflectionProperty::IS_PUBLIC
+        private int $propertyVisibility = ReflectionProperty::IS_PRIVATE | ReflectionProperty::IS_PROTECTED | ReflectionProperty::IS_PUBLIC
     ) {
         $this->dataSetProvided = $this->object instanceof DataSetInterface;
-        $this->cacheKey = get_class($this->object) . '_' . $this->propertyVisibility;
-    }
-
-    public function getRules(): iterable
-    {
-        $objectHasRules = $this->object instanceof RulesProviderInterface;
-        $rules = $objectHasRules ? $this->object->getRules() : [];
-
-        // Providing data set assumes object has its own attributes and rules getting logic. So further parsing of
-        // Reflection properties and rules is skipped intentionally.
-        if ($this->dataSetProvided || $objectHasRules === true) {
-            return $rules;
-        }
-
-        if ($this->hasCachedRules()) {
-            return $this->getCachedRules();
-        }
-
-        foreach ($this->getReflectionProperties() as $property) {
-            $attributes = $property->getAttributes(RuleInterface::class, ReflectionAttribute::IS_INSTANCEOF);
-            foreach ($attributes as $attribute) {
-                $rule = $attribute->newInstance();
-                $rules[$property->getName()][] = $rule;
-
-                if ($rule instanceof AttributeEventInterface) {
-                    $rule->afterInitAttribute($this);
-                }
-            }
-        }
-
-        $this->setCachedRules($rules);
-
-        return $rules;
-    }
-
-    /**
-     * @return ReflectionProperty[] Used to avoid error "Typed property must not be accessed before initialization".
-     */
-    private function getReflectionProperties(): array
-    {
-        if ($this->reflectionProperties !== null) {
-            return $this->reflectionProperties;
-        }
-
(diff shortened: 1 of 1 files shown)


src/DataSet/ObjectDataSet.php
```diff
@@ -13,106 +13,48 @@
...
+
+    private static array $cache = [];
+    private ?string $cacheKey;
 
     public function __construct(
         private object $object,
-        private int $propertyVisibility = ReflectionProperty::IS_PRIVATE |
-        ReflectionProperty::IS_PROTECTED |
-        ReflectionProperty::IS_PUBLIC
+        private int $propertyVisibility = ReflectionProperty::IS_PRIVATE | ReflectionProperty::IS_PROTECTED | ReflectionProperty::IS_PUBLIC
```
arogachev, line 35: It does not fix max line length this way.

src/DataSet/ObjectDataSet.php
```diff
@@ -13,106 +13,48 @@
 use Yiisoft\Validator\RulesProviderInterface;
 
 use function array_key_exists;
-use function get_class;
 
 /**
  * This data set makes use of attributes introduced in PHP 8. It simplifies rules configuration process, especially for
- * nested data and relations. Please refer to the guide for examples.
+ * nested data and relations. Please refer to the guide for example.
```
arogachev, line 19: ?

arogachev: Actually currently there are no examples with `ObjectDataSet` in README at all.

src/DataSet/ObjectDataSet.php
```diff
@@ -13,106 +13,48 @@
...
-    private ?array $reflectionProperties = null;
-    private ?array $data = null;
+    private bool $rulesProvided;
 
-    private static array $cachedRules = [];
-    private string $cacheKey;
+    private ?array $cachedData = null;
+
+    private static array $cache = [];
+    private ?string $cacheKey;
```
arogachev, line 31: Why it's nullable?

src/DataSet/ObjectDataSet.php
```diff
@@ -13,106 +13,48 @@
...
 
-        $this->reflectionProperties = $reflectionProperties;
-
-        return $reflectionProperties;
-    }
-
-    public function getObject(): object
-    {
-        return $this->object;
+        return $this->getCachedRules();
```
arogachev, line 57: In my opinion here should be this sequence:

1. Attempt to get rules from cache. If they are found - return cached version.
2. Actual rules getting logic.
3. Update cache and return rules.

Similar with reflection properties.

src/DataSet/ObjectDataSet.php
```diff
@@ -137,32 +79,65 @@ public function getData(): array
...
+    private function getCachedReflectionProperties(): array
     {
-        return self::$cachedRules[$this->cacheKey];
+        $this->assertRulesAndReflectionPropertiesInCache();
+        return self::$cache[$this->cacheKey]['reflectionProperties'];
     }
 
-    private function setCachedRules(array $rules): void
+    // TODO: use Generator to collect attributes
+    private function assertRulesAndReflectionPropertiesInCache(): void
```
arogachev, line 107: Too much is going on here. And technically it's not assert. The purpose of assert is just verify the presence of data in cache in this case.

src/DataSet/ObjectDataSet.php
```diff
@@ -137,32 +79,65 @@ public function getData(): array
...
+                $rule = $attribute->newInstance();
+                $rules[$property->getName()][] = $rule;
+
+                if ($rule instanceof AttributeEventInterface) {
+                    $rule->afterInitAttribute($this);
+                }
+            }
+        }
+
+        self::$cache[$this->cacheKey] = [
```
arogachev, line 138: Suggest to restore cache shortcuts and structure (`ArrayShape`).

src/DataSet/ObjectDataSet.php
```diff
@@ -137,32 +79,65 @@ public function getData(): array
...
     {
-        self::$cachedRules[$this->cacheKey] = $rules;
+        if (array_key_exists($this->cacheKey, self::$cache)) {
+            return;
+        }
+
+        $rules = [];
+        $reflectionProperties = [];
+
+        $reflection = new ReflectionObject($this->object);
```
arogachev, line 116: Suggest to split / restore how it was in original PR.

src/DataSet/ObjectDataSet.php
```diff
@@ -13,106 +13,48 @@
...
-    /**
-     * @var ReflectionProperty[]
-     */
-    private ?array $reflectionProperties = null;
-    private ?array $data = null;
+    private bool $rulesProvided;
 
-    private static array $cachedRules = [];
-    private string $cacheKey;
+    private ?array $cachedData = null;
```
arogachev, line 28: Maybe emphasize that is instance data. Also to avoid confusion with complex data in static cache.

arogachev (changes requested):
:heavy_check_mark:

src/DataSet/ObjectDataSet.php
vjik: done

src/DataSet/ObjectDataSet.php
vjik: done

src/DataSet/ObjectDataSet.php
vjik: fixed

src/DataSet/ObjectDataSet.php
vjik: `$instanceDataCache` ?

src/DataSet/ObjectDataSet.php
vjik: If we do this, it will take several passes through the properties (foreach for reflection properties). In the current implementation we use foreach once.

src/DataSet/ObjectDataSet.php
vjik: Goal - check cache and if empty, then get and cache data. Agree that need to rename method. Can you suggest name?

arogachev: I still think that renaming method is not enough in this case.

src/DataSet/ObjectDataSet.php
arogachev: Still actual.

src/DataSet/ObjectDataSet.php
arogachev: Still actual.

src/DataSet/ObjectDataSet.php
arogachev: Discussed it in PM. Yes, an additional pass is required, but on the cached data. And with the amount of properties the profit will likely be miserable. But separation is more organized and readable ("design first, optimize later" principle).

Issue #327, opened 2022-10-05
Add support for non-static methods in `Callback::$method`


arogachev:
- #321
- https://github.com/yiisoft/form/blob/master/src/FormModel.php#L318-L321

PR #343, opened 2022-10-15
Improve callable rule

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fix | #327 

- [x] Added ability to run non-static methods
- [x] Static methods are still supported


tests/DataSet/ObjectDataSetTest.php
```diff
@@ -259,23 +257,17 @@ public function testValidateWithCallbackMethod(): void
...
-        return [
-            [new ObjectWithNonExistingCallbackMethod()],
-            [new ObjectWithNonPublicCallbackMethod()],
-            [new ObjectWithNonStaticCallbackMethod()],
-        ];
-    }
-
-    /**
-     * @link https://github.com/yiisoft/validator/issues/223
-     * @dataProvider validateWithWrongCallbackMethodDataProvider
```
vjik, line 273: Why removed this tests?

tests/Rule/CallbackTest.php
```diff
@@ -125,6 +126,35 @@ public function testValidateUsingMethodOutsideAttributeScope(): void
         $validator->validate(null, [$rule]);
     }
 
+    public function testValidateUsingNonStaticMethod(): void
```
vjik, line 129: Here we can don't create separate test method. Use `dataValidationFailed()`.

src/Rule/Callback.php
```diff
@@ -79,11 +79,19 @@ public function afterInitAttribute(DataSetInterface $dataSet): void
...
 
-        try {
-            $this->callback = Closure::fromCallable([$dataSet->getObject()::class, $this->method]);
-        } catch (TypeError) {
-            throw new InvalidArgumentException('Method must exist and have public and static modifiers.');
+        $object = $dataSet->getObject();
+        $method = $this->method;
+
+        $reflection = new ReflectionObject($object);
+        if (!$reflection->hasMethod($method)) {
```
vjik, line 86: Here `$method` can be null and this case throw exception.

vjik (changes requested):

tests/DataSet/ObjectDataSetTest.php
xepozz: Because the functionality it implements became as a feature

tests/DataSet/ObjectDataSetTest.php
```diff
@@ -16,8 +16,6 @@
 use Yiisoft\Validator\RuleInterface;
 use Yiisoft\Validator\Tests\Support\Data\ObjectWithCallbackMethod\ObjectWithCallbackMethod;
 use Yiisoft\Validator\Tests\Support\Data\ObjectWithCallbackMethod\ObjectWithNonExistingCallbackMethod;
-use Yiisoft\Validator\Tests\Support\Data\ObjectWithCallbackMethod\ObjectWithNonPublicCallbackMethod;
```
arogachev, line 19: Need to delete according data files as well.

tests/Rule/CallbackTest.php
```diff
@@ -101,6 +102,29 @@ public function dataValidationFailed(): array
...
+                    public function __construct(
+                        #[Callback(method: 'validateName')]
+                        private $age,
+                    ) {
+                    }
+
+                    private function validateName($value, $rule, $context): Result
+                    {
+                        if ($value !== $this->age) {
+                            throw new RuntimeException('Method scope was not bound the object.');
```
arogachev, line 116:
```suggestion
                            throw new RuntimeException('Method scope was not bound with the object.');
```

src/Rule/Callback.php
```diff
@@ -79,11 +79,23 @@ public function afterInitAttribute(DataSetInterface $dataSet): void
             return;
         }
 
-        try {
-            $this->callback = Closure::fromCallable([$dataSet->getObject()::class, $this->method]);
-        } catch (TypeError) {
-            throw new InvalidArgumentException('Method must exist and have public and static modifiers.');
+        if ($this->method === null) {
+            return;
```
arogachev, line 83: A test is needed for a case when no method was passed.

tests/Rule/CallbackTest.php
```diff
@@ -101,6 +102,29 @@ public function dataValidationFailed(): array
...
             ],
+            'non-static callable' => [
+                new class (1) {
+                    public function __construct(
+                        #[Callback(method: 'validateName')]
+                        private $age,
+                    ) {
+                    }
+
+                    private function validateName($value, $rule, $context): Result
```
arogachev, line 113:
```suggestion
                    private function validateName(mixed $value, RuleInterface $rule, ValidationContext $context): Result
```

tests/DataSet/ObjectDataSetTest.php
```diff
@@ -259,23 +257,17 @@ public function testValidateWithCallbackMethod(): void
         $this->assertSame(['name' => ['Value must be "foo"!']], $result->getErrorMessagesIndexedByPath());
     }
 
-    public function validateWithWrongCallbackMethodDataProvider(): array
-    {
-        return [
-            [new ObjectWithNonExistingCallbackMethod()],
-            [new ObjectWithNonPublicCallbackMethod()],
```
arogachev, line 266: The test for public method is still needed.

src/Rule/Callback.php
```diff
@@ -79,11 +79,23 @@ public function afterInitAttribute(DataSetInterface $dataSet): void
             return;
         }
 
-        try {
-            $this->callback = Closure::fromCallable([$dataSet->getObject()::class, $this->method]);
-        } catch (TypeError) {
-            throw new InvalidArgumentException('Method must exist and have public and static modifiers.');
```
arogachev, line 85: The check for non-public method is still needed.

xepozz: If it works for private methods, public and protected methods should work also.

src/Rule/Callback.php
xepozz: It is already present https://github.com/yiisoft/validator/blob/72cf6616a62d3faa87c1b6897741a9d7e4c97d59/tests/Rule/CallbackTest.php#L61

src/Rule/Callback.php
arogachev: Missed that it was already presented. Explicit `public` and `protected` tests won't be redundant in my opinion. But yes, maybe private is enough for now.

src/Rule/Callback.php
arogachev: Actually, let's leave it for later. There are other non-covered parts in Callback rule (not related with this PR's diff) and we have a task of raising code coverage close to 100% before 1.0.0 release anyway.

tests/Rule/CallbackTest.php
```diff
@@ -101,6 +103,42 @@ public function dataValidationFailed(): array
...
+                            throw new RuntimeException('Method scope was not bound to the object.');
+                        }
+
+                        $result = new Result();
+                        $result->addError('Hello from non-static method.');
+
+                        return $result;
+                    }
+
+                    private static function staticValidateName(mixed $value, RuleInterface $rule, ValidationContext $context): Result
```
arogachev, line 127:
```suggestion
                    private static function staticValidateName(
                        mixed $value, RuleInterface $rule, 
                        ValidationContext $context
                    ): Result
```

arogachev: This line is too long.

arogachev (approved):

PR #344, opened 2022-10-16
Fix rules serialization

xepozz:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌


src/RulesDumper.php
```diff
@@ -54,16 +53,17 @@ private function fetchOptions(iterable $rules): array
...
+            if (is_iterable($rule)) {
                 $result[$attribute] = $this->fetchOptions($rule);
             } elseif ($rule instanceof SerializableRuleInterface) {
                 $result[$attribute] = array_merge([$rule->getName()], $rule->getOptions());
             } elseif ($rule instanceof RuleInterface) {
                 $result[$attribute] = [$rule->getName()];
             } else {
                 throw new InvalidArgumentException(sprintf(
-                    'Rules should be a rule or an array of rules that implements %s.',
+                    'Rules should be a rule or an array of rules that implements "%s". Type "%s" given.',
```
arogachev, line 64:
```suggestion
                    'A `$rules` array item must be a single rule or an iterable of rules. Each rule must implement "%s". Type "%s" given.',
```

arogachev:
Could you also add a test?

arogachev (changes requested):
:+1:

PR #392, opened 2022-11-25
Fix messages

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | -


arogachev (changes requested):
Was thinking about that too. Even I don't quite like the additional nesting with repeated names, seems like it's better for consistency.

Regarding to repeated names (`$options['message']['message']` or `$options['someOtherMessage']['message']`), maybe change the inner name to something like `text`?

```php
'message' => [ // Or "someOtherMessage"
    'text' => '...', // Instead of "message"
    'parameters' => [], // Optional
];
```

Also there are more places to change, for example:

https://github.com/yiisoft/validator/blob/d75c576bf886b967fe4dbf42641d5c60939962c4/src/Rule/AtLeast.php#L87

samdark (approved):

arogachev (approved):

Issue #358, opened 2022-10-23
Actualize docs and README


arogachev:
- [x] `README.md` @arogachev
- [x] `src/AfterInitAttributeEventInterface.php` @arogachev
- [x] `src/DataSet/ArrayDataSet.php` @arogachev
- [x] `src/DataSet/ObjectDataSet.php` @arogachev
- [x] `src/DataSet/SingleValueDataSet.php` @arogachev
- [x] `src/DataSetInterface.php` @arogachev
- [x] `src/EmptyCriteria/NeverEmpty.php` @arogachev
- [x] `src/EmptyCriteria/WhenEmpty.php` @arogachev
- [x] `src/EmptyCriteria/WhenNull.php` @arogachev
- [x] `src/Error.php` @arogachev, @samdark #487
- [x] `src/Exception/InvalidCallbackReturnTypeException.php` @arogachev
- [x] `src/Exception/RuleHandlerInterfaceNotImplementedException.php` @arogachev
- [x] `src/Exception/RuleHandlerNotFoundException.php` @arogachev
- [x] `src/Exception/UnexpectedRuleException.php` @arogachev
- [x] `src/Helper/DataSetNormalizer.php` @arogachev
- [x] `src/Helper/ObjectParser.php` @arogachev
- [x] `src/Helper/RulesNormalizer.php` @arogachev #506
- [x] `src/Helper/SkipOnEmptyNormalizer.php` @arogachev
- [x] `src/LimitInterface.php` @arogachev
- [x] `src/PostValidationHookInterface.php` @arogachev
- [x] `src/PropagateOptionsInterface.php` @arogachev
- [x] `src/Result.php` @samdark #487
- [x] `src/Rule/AtLeast.php`@samdark #489
- [x] `src/Rule/AtLeastHandler.php` @samdark #489
- [x] `src/Rule/Boolean.php` @arogachev #485
- [x] `src/Rule/BooleanHandler.php` @arogachev #485 
- [x] `src/Rule/Callback.php` @samdark #489
- [x] `src/Rule/CallbackHandler.php` @samdark #489
- [x] `src/Rule/Compare.php` @samdark #489
- [x] `src/Rule/CompareHandler.php` @samdark #489
- [x] `src/Rule/CompareTo.php` @samdark #489
- [x] `src/Rule/Composite.php` @arogachev #496
- [x] `src/Rule/CompositeHandler.php` @arogachev #496
- [x] `src/Rule/Count.php` @samdark #489
[...]

PR #402, opened 2022-11-29
Docs, part 1

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #358 

- Add custom Psalm type `WhenType` and apply it across all project.
- Rename `SerializableRuleInterface` to `RuleWithOptionsInterface` because technically `RuleInterface` is serializable too but with name only.
- Add docs for `RuleInterface` and `RuleWithOptionsInterface`.


src/RuleWithOptionsInterface.php
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * ```php
+     * [
+     *     'property1' => $this->>property1,
+     *     'property2' => $this->>property2,
+     *     // ...
+     * ];
+     * ```
+     *
+     * Sometimes it's helpful to include dynamic options as well:
```
vjik, line 34: What is dynamic options?

src/RuleWithOptionsInterface.php
```diff
@@ -0,0 +1,64 @@
...
+     * ```php
+     * 'message' => [ // Array is used with no parameters too.
+     *     'template => 'A message without parameters.',
+     *     'parameters' => [], // Explicitly specified even for empty parameters.
+     * ],
+     * 'anotherMessage' => [
+     *     'template' => 'A message with {parameter}.', // Handler dependent parameters
+     *     'parameters' => ['property4' => $this->property4], // Property dependent parameters
+     * ],
+     * 'wrongMessage' => 'Wrong message.' // This is possible, but wrong. Use example above for consistent structure.
```
vjik, line 55: `This is possible` — suggest remove this. By our agreements messages in options always must have structure:

```
'message' => ['template' => ..., 'parameters' => [...]]
```

src/RuleWithOptionsInterface.php
arogachev: For example:

https://github.com/yiisoft/validator/blob/f4e4de303fb05877af5aa6716d04dce579939cca/src/Rule/Compare.php#L169-L172

https://github.com/yiisoft/validator/blob/f4e4de303fb05877af5aa6716d04dce579939cca/src/Rule/Compare.php#L138-L148

The idea was to show that not only properties can be included as options.

src/RuleWithOptionsInterface.php
arogachev: Fixed. Just in case, you mean remove the comment part, right? Or remove this example completely?

vjik: Now is fine!

src/RuleWithOptionsInterface.php
vjik: May be completely remove this block? It's base PHP availabilities.

arogachev: OK. Removed it completely.

vjik (approved):

PR #403, opened 2022-11-29
Allow pass `callable` as rules to `Validator::validate()` + Minor fix rules preparing in `Nested` rule

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | -

### DIFF
--- a/src/Helper/RulesNormalizer.php
+++ b/src/Helper/RulesNormalizer.php
@@ -28,7 +28,7 @@ final class RulesNormalizer
      * @return iterable<int|string, iterable<int|string, RuleInterface>>
      */
     public static function normalize(
-        iterable|object|null $rules,
+        iterable|object|callable|null $rules,
         mixed $data = null,
         ?callable $defaultSkipOnEmptyCriteria = null,
     ): iterable {
@@ -65,7 +65,7 @@ final class RulesNormalizer
      * @throws InvalidArgumentException
      */
     private static function prepareRulesArray(
-        iterable|object|null $rules,
+        iterable|object|callable|null $rules,
         mixed $data,
     ): iterable {
         if ($rules === null) {
@@ -80,6 +80,10 @@ final class RulesNormalizer
 
         if ($rules instanceof RuleInterface) {
             return [$rules];
+        }
+
+        if (is_callable($rules)) {
+            return [new Callback($rules)];
         }
 
         /** @psalm-suppress RedundantConditionGivenDocblockType */
--- a/src/Helper/RulesNormalizer.php
+++ b/src/Helper/RulesNormalizer.php
@@ -78,12 +78,8 @@ final class RulesNormalizer
             return $rules->getRules();
         }
 
-        if ($rules instanceof RuleInterface) {
+        if ($rules instanceof RuleInterface || is_callable($rules)) {
             return [$rules];
-        }
-
-        if (is_callable($rules)) {
-            return [new Callback($rules)];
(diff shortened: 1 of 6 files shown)


src/Helper/RulesNormalizer.php
```diff
@@ -82,6 +82,10 @@ private static function prepareRulesArray(
             return [$rules];
         }
 
+        if (is_callable($rules)) {
```
arogachev, line 85: Maybe modify the condition above instead?

```php
if ($rules instanceof RuleInterface || $is_callable($rules)) {
    return [$rules];
}
```

And the `Callback` conversion is already performed here for single rule:

https://github.com/yiisoft/validator/blob/9ae20f808ddd706ac3cfb9ac74c8c11af1123292/src/Helper/RulesNormalizer.php#L115-L117

tests/ValidatorTest.php
```diff
@@ -186,6 +186,11 @@ public function getRules(): iterable
                 [],
                 new Foo(),
             ],
+            'array-and-callable' => [
+                ['' => ['test message']],
+                [],
+                static fn () => (new Result())->addError('test message'),
```
arogachev, line 192:
```suggestion
                static fn (): Result => (new Result())->addError('test message'),
```

src/Rule/Nested.php
```diff
@@ -167,10 +160,14 @@ private function prepareRules(iterable|object|string|null $source): void
...
             $rules = $source->getRules();
-        } elseif (!$source instanceof Traversable && !is_array($source)) {
+        } elseif (is_string($source) && class_exists($source)) {
             $rules = (new AttributesRulesProvider($source, $this->rulesPropertyVisibility))->getRules();
-        } else {
+        } elseif (is_iterable($source)) {
             $rules = $source;
+        } else {
+            throw new InvalidArgumentException(
+                'Invalid rules object.'
```
arogachev, line 169: It's not necessarily an object. Can be a simple string or a string with not existing class name.

src/Helper/RulesNormalizer.php
vjik: done

src/Rule/Nested.php
arogachev: Also it's a bit broad - for a user it can be hard to understand what exactly is wrong. We can have more specific exceptions with more clear messages, for example a separate exception for string type - that only existing class names are expected. But it can be improved later, with a friendly exception maybe.

arogachev (changes requested):
:+1:

src/Rule/Nested.php
```diff
@@ -167,10 +160,14 @@ private function prepareRules(iterable|object|string|null $source): void
...
+            );
```
arogachev, line 170:
```suggestion
            $message = 'The $rules argument passed to Nested rule can be either: a null, an object implementing ' .
                'RulesProviderInterface, a class string or an iterable.';

            throw new InvalidArgumentException($message);
```

src/Rule/Nested.php
arogachev: > Please, suggest message. I'm replace.

https://github.com/yiisoft/validator/pull/403/files#r1034697378

src/Rule/Nested.php
vjik: done

PR #408, opened 2022-11-30
Docs, part 2

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #358

- Add docs for `SkipOnEmptyInterface` and `SkipOnEmptyTrait`.
- Fix ambiguity in `SkipOnEmptyTrait::getSkipOnEmptyOption()` when handling callable.


src/SkipOnEmptyInterface.php
```diff
@@ -4,9 +4,96 @@
...
+     *     return $this->skipOnEmpty;
+     * }
+     * ```
+     *
+     * During pre-validation phase it will be normalized to an "empty criteria" - a callable identifying when and which
+     * values exactly must be considered as empty and skipped or not skipped at all.
+     *
+     * @return bool|callable|null A raw value set in the constructor:
+     *
+     * - `null` and `false` - never skip a rule (the validated value is always considered as not empty). Matching
```
vjik, line 65: Null - will be used default empty criteria from validator.

vjik (changes requested):

src/SkipOnEmptyInterface.php
```diff
@@ -4,9 +4,96 @@
...
 namespace Yiisoft\Validator;
 
+use Yiisoft\Validator\EmptyCriteria\NeverEmpty;
+use Yiisoft\Validator\EmptyCriteria\WhenEmpty;
+use Yiisoft\Validator\EmptyCriteria\WhenNull;
+use Yiisoft\Validator\Rule\Trait\SkipOnEmptyTrait;
+
+/**
+ * An optional interface for rules to implement. It allows skipping validation for a rule when the validated value is
+ * "empty". Requires an additional private class property `$skipOnEmpty`. In package rules it's `null` by default:
```
samdark, line 14: > Requires an additional private class property `$skipOnEmpty`.

Interface can't require a property, especially private one.

samdark: I think that belongs to `SkipOnEmptyTrait` description and not here.

src/SkipOnEmptyInterface.php
```diff
@@ -4,9 +4,96 @@
...
+ * }
+ * ```
+ *
+ * The package ships with {@see SkipOnEmptyTrait} which already implements that interface. All you have to do is include
+ * it in the rule class.
+ */
 interface SkipOnEmptyInterface
 {
+    /**
+     * A setter to change `$skipOnEmpty` property. Must be implemented following immutability concept:
```
samdark, line 31: Interface knows nothing about how the value is stored so we should not mention implementation details.

samdark: Immuntability could be mentioned as a requirement, but I don't think an example is needed.

src/SkipOnEmptyInterface.php
```diff
@@ -4,9 +4,96 @@
...
+     * ```
+     *
+     * @param bool|callable|null $value
+     *
+     * @return $this
+     */
     public function skipOnEmpty(bool|callable|null $value): static;
 
+    /**
+     * A getter for `$skipOnEmpty` property. Returned value is unchanged. Typical implementation is just a simple
```
samdark, line 50: Same here. Should not mention properties. That's not part of the interface.

samdark (changes requested):

arogachev:
The part with default value in `SkipOnEmptyTrait::getSkipOnEmptyOption()` can be handled later - #411.

vjik (approved):

samdark (approved):

PR #418, opened 2022-12-02
Docs, part 4

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #358 

- Add docs to `WhenInterface` and `WhenTrait`.

### DIFF
--- a/src/Rule/Trait/WhenTrait.php
+++ b/src/Rule/Trait/WhenTrait.php
@@ -8,12 +8,31 @@ use Closure;
 use Yiisoft\Validator\WhenInterface;
 
 /**
+ * An implementation for {@see WhenInterface} intended to be included in rules. Requires an additional private class
+ * property `$when`. In package rules it's `null` by default:
+ *
+ * ```php
+ * public function __construct(
+ *     // ...
+ *     private Closure|null $when = null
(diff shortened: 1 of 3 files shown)


src/Rule/Trait/WhenTrait.php
```diff
@@ -23,7 +42,12 @@ public function when(Closure|null $value): static
     }
 
     /**
-     * @psalm-return WhenType
+     * A getter for `$when` property.
+     *
+     * @psalm-return WhenType Current value:
```
vjik, line 47: For correct work documentation generators we can make base phpdoc annotations (`@return`, `@param`, ...) with description and psalm annotations without description. For example:

```
     *
     * @return Closure|null Current value:
     * 
     * - `null` - always apply the validation.
     * - `callable` - apply the validation depending on a return value: `true` - apply, `false` - do not apply.
     *
     * @psalm-return WhenType
     *
````

`@return` for humans and generators.

`@psalm-return` for psalm only.

vjik (changes requested):

samdark (approved):

PR #421, opened 2022-12-05
Docs, part 5

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #358

- Add docs to `LimitInterface` and `LimitTrait`.

### DIFF
--- a/src/LimitInterface.php
+++ b/src/LimitInterface.php
@@ -32,7 +32,8 @@ interface LimitInterface
     public function getMax(): int|null;
 
     /**
-     * Returns current "exactly" value. A shortcut for the case when {@see getMin()} and {@see getMax()} have the same
+     * Returns current "exactly" value meant for strict comparison.
+     * It is a shortcut for the case when {@see getMin()} and {@see getMax()} have the same
      * not null value.
      *
      * @return int|null A number representing "exactly" value. `null` means no strict comparison so lower / upper limits
--- a/src/LimitInterface.php
+++ b/src/LimitInterface.php
@@ -5,10 +5,10 @@ declare(strict_types=1);
 namespace Yiisoft\Validator;
 
 /**
- * An optional interface for rules to implement. Rules including it must have minimum and maximum limits and also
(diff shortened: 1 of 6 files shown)


vjik (approved):

src/LimitInterface.php
```diff
@@ -4,17 +4,62 @@
 
 namespace Yiisoft\Validator;
 
+/**
+ * An optional interface for rules to implement. Rules including it must have minimum and maximum limits and also
+ * "exactly" measure for strict equality. Boundaries can reflect countable items as well as absolute values.
+ *
+ * The package ships with {@see LimitTrait} which already implements that interface. All you have to do is include it in
+ * the rule class along with the interface.
```
samdark, line 12:
```suggestion
 * An optional interface for rules to implement. Rules implementing it must have minimum and maximum limits and also
 * "exactly" value for strict equality. Boundaries can reflect countable items as well as absolute values.
 *
 * The package ships with {@see LimitTrait} which already implements that interface. All you have to do is to include it in
 * the rule class along with the interface.
```

src/LimitInterface.php
```diff
@@ -4,17 +4,62 @@
...
 
-    public function getExactly(): ?int;
+    /**
+     * Returns current "exactly" value. A shortcut for the case when {@see getMin()} and {@see getMax()} have the same
+     * not null value.
+     *
+     * @return int|null A number representing "exactly" value. `null` means no strict comparison so lower / upper limits
+     * / both must be set.
+     *
+     * @see getNotExactlyMessage() for related error message
```
samdark, line 41:
```suggestion
     * Returns current "exactly" value meant for strict comparison.
     * It is a shortcut for the case when {@see getMin()} and {@see getMax()} have the same
     * not null value.
     *
     * @return int|null A number representing "exactly" value. `null` means no strict comparison so lower / upper limits
     * / both must be set.
     *
     * @see getNotExactlyMessage() for related error message
```

samdark (approved):

Issue #382, opened 2022-11-21
Integrate limit interface and traits into `Number` rule


arogachev:

arogachev:
Working on it.

PR #422, opened 2022-12-06
Apply limit interface and traits to `Number` rule

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #382


vjik (commented):
I think not need use limit interface in `Number` rule.

1) More difficult trait: added float values + added `allowNegativeLimits` and `requireLimits` options.

2) In `Number` added overhead option `exactly`.

src/Rule/Trait/LimitTrait.php
```diff
@@ -89,21 +89,23 @@ trait LimitTrait
...
-        int|null $max,
-        int|null $exactly,
+        int|float|null $min,
+        int|float|null $max,
+        int|float|null $exactly,
         string $lessThanMinMessage,
         string $greaterThanMinMessage,
-        string $notExactlyMessage
+        string $notExactlyMessage,
+        bool $requireLimits = true,
```
arogachev, line 107: Need to add PHPDoc with description.

samdark:
$equal = \abs($a - $b) < PHP_FLOAT_EPSILON;

Should be used for floats.

vjik:
A little discussed. Idea: make improvements of `Number` (names, floats), but don't change  `LimitTrait` and don't use it into `Number`.

PR #435, opened 2022-12-07
Docs, part 7

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #358

### DIFF
--- a/src/DataSet/ArrayDataSet.php
+++ b/src/DataSet/ArrayDataSet.php
@@ -11,7 +11,7 @@ use Yiisoft\Validator\Helper\DataSetNormalizer;
 use function array_key_exists;
 
 /**
- * A data set for storing data as associative array, where keys are attribute names and values are their corresponding
+ * A data set for storing data as an associative array, where keys are attribute names and values are their corresponding
  * values. An example of usage:
  *
  * ```php
--- a/src/DataSet/ArrayDataSet.php
+++ b/src/DataSet/ArrayDataSet.php
@@ -18,7 +18,7 @@ use function array_key_exists;
  * $dataSet = new ArrayDataSet(['name' => 'John', 'age' => 18]);
  * ```
  *
- * When using with validator, there is no need to do it manually, because array will be automatically wrapped with
+ * When using validator, there is no need to wrap your data manually. Array will be automatically wrapped with
  * {@see ArrayDataSet} by {@see DataSetNormalizer} during validation.
  */
 final class ArrayDataSet implements DataSetInterface
--- a/src/DataSet/ArrayDataSet.php
+++ b/src/DataSet/ArrayDataSet.php
@@ -11,8 +11,8 @@ use Yiisoft\Validator\Helper\DataSetNormalizer;
 use function array_key_exists;
 
 /**
- * A data set for storing data as an associative array, where keys are attribute names and values are their corresponding
- * values. An example of usage:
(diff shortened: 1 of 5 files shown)


src/DataSet/ArrayDataSet.php
```diff
@@ -6,24 +6,64 @@
...
 
+/**
+ * A data set for storing data as associative array, where keys are attribute names and values are their corresponding
+ * values. An example of usage:
+ *
+ * ```php
+ * $dataSet = new ArrayDataSet(['name' => 'John', 'age' => 18]);
+ * ```
+ *
+ * When using with validator, there is no need to do it manually, because array will be automatically wrapped with
```
samdark, line 21:
```suggestion
 * When using validator, there is no need to wrap your data manually. Array will be automatically wrapped with
```

src/DataSet/ArrayDataSet.php
```diff
@@ -6,24 +6,64 @@
 
 use Yiisoft\Validator\DataSetInterface;
 
+use Yiisoft\Validator\Helper\DataSetNormalizer;
+
 use function array_key_exists;
 
+/**
+ * A data set for storing data as associative array, where keys are attribute names and values are their corresponding
+ * values. An example of usage:
```
samdark, line 15:
```suggestion
 * A data set for storing data as an associative array, where keys are attribute names and values are their corresponding
 * values. An example of usage:
```

src/DataSet/ArrayDataSet.php
```diff
@@ -6,24 +6,64 @@
...
+         * @var array A mapping between attribute names and their values.
+         */
+        private array $data = [],
+    ) {
     }
 
+    /**
+     * Returns an attribute value by its name.
+     *
+     * Note that in case of non-existing attribute a default `null` value is returned. If you need to just check the
```
samdark, line 37:
```suggestion
     * Note that in case of non-existing attribute a default `null` value is returned. If you need to check the
```

src/DataSet/SingleValueDataSet.php
```diff
@@ -7,24 +7,64 @@
...
+ * ```php
+ * $dataSet = new SingleValueDataSet(5);
+ * $dataSet = new SingleValueDataSet(2.5);
+ * $dataSet = new SingleValueDataSet('text');
+ * $dataSet = new SingleValueDataSet(null);
+ * $dataSet = new SingleValueDataSet(false);
+ * ```
+ *
+ * When using with validator, there is no need to do it manually, because such types be automatically wrapped with
+ * {@see SingleValueDataSet} by {@see DataSetNormalizer} during validation.
```
samdark, line 23:
```suggestion
 * When using validator, there is no need to wrap data manually. Such types be automatically wrapped with
 * {@see SingleValueDataSet} by {@see DataSetNormalizer} during validation.
```

src/DataSetInterface.php
```diff
@@ -4,17 +4,34 @@
 
 namespace Yiisoft\Validator;
 
+/**
+ * An interface for unifying access to different types of validated data. A class implementing it called "data set".
```
samdark, line 8:
```suggestion
 * An interface for unifying access to different types of validated data. A class implementing it is called "data set".
```

samdark (changes requested):

PR #446, opened 2022-12-09
Allow use `Each`, `Callback` and `StopOnError` as class attribute

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | -

### DIFF
--- a/src/Rule/Callback.php
+++ b/src/Rule/Callback.php
@@ -32,7 +32,7 @@ final class Callback implements
     use SkipOnErrorTrait;
     use WhenTrait;
 
-    private ?object $objectBeingValidated = null;
+    private ?object $objectValidated = null;
 
     public function __construct(
         /**
@@ -75,15 +75,15 @@ final class Callback implements
         return $this->method;
     }
 
-    public function getObjectBeingValidated(): ?object
+    public function getObjectValidated(): ?object
     {
-        return $this->objectBeingValidated;
+        return $this->objectValidated;
     }
 
     public function afterInitAttribute(object $object, int $target): void
     {
         if ($target === Attribute::TARGET_CLASS) {
-            $this->objectBeingValidated = $object;
+            $this->objectValidated = $object;
         }
 
         if ($this->method === null) {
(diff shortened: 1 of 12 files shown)


src/Rule/Callback.php
```diff
@@ -32,6 +32,8 @@ final class Callback implements
     use SkipOnErrorTrait;
     use WhenTrait;
 
+    private ?object $objectBeingValidated = null;
```
samdark, line 35:
```suggestion
    private ?object $objectValidated = null;
```

src/Rule/Callback.php
```diff
@@ -73,8 +75,17 @@ public function getMethod(): string|null
...
 
+    public function getObjectBeingValidated(): ?object
+    {
+        return $this->objectBeingValidated;
+    }
+
     public function afterInitAttribute(object $object, int $target): void
     {
+        if ($target === Attribute::TARGET_CLASS) {
+            $this->objectBeingValidated = $object;
```
samdark, line 86:
```suggestion
            $this->objectValidated = $object;
```

src/Rule/Callback.php
```diff
@@ -73,8 +75,17 @@ public function getMethod(): string|null
         return $this->method;
     }
 
+    public function getObjectBeingValidated(): ?object
+    {
+        return $this->objectBeingValidated;
```
samdark, line 80:
```suggestion
        return $this->objectValidated;
```

src/Rule/CallbackHandler.php
```diff
@@ -27,7 +27,7 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
             throw new InvalidArgumentException('Using method outside of attribute scope is prohibited.');
         }
 
-        $result = $callback($value, $rule, $context);
+        $result = $callback($rule->getObjectBeingValidated() ?? $value, $rule, $context);
```
samdark, line 30:
```suggestion
        $result = $callback($rule->getObjectValidated() ?? $value, $rule, $context);
```

samdark (changes requested):

vjik:
@samdark I'm renamed the suggested properties/methods and rename same properties/merthods in `Count` rule.

samdark (approved):

PR #464, opened 2022-12-14
Add `DataWrapperInteface`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  |


samdark (approved):

src/Rule/Count.php
```diff
@@ -158,11 +156,4 @@ public function getHandler(): string
     {
         return CountHandler::class;
     }
-
-    public function afterInitAttribute(object $object, int $target): void
```
samdark, line 162: Is it alright?

samdark:
Diff should be re-checked.

src/Rule/Count.php
vjik: Yes. Now object itself passed to validation when rule applied to whole object.

vjik:
> Diff should be re-checked.

Yes. Merge with master is done wrong.

samdark:
Resolved conflicts, added docs.

PR #473, opened 2022-12-19
Add more phpdoc documentation

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #358

### DIFF
--- a/src/AttributeTranslator/ArrayAttributeTranslator.php
+++ b/src/AttributeTranslator/ArrayAttributeTranslator.php
@@ -11,6 +11,10 @@ use Yiisoft\Validator\AttributeTranslatorInterface;
  */
 final class ArrayAttributeTranslator implements AttributeTranslatorInterface
 {
+    /**
+     * @param array $translations Translations array where each key is an attribute name and the corresponding name is
+     * a translation.
+     */
     public function __construct(
         /**
          * @var array<string,string>
--- a/src/AttributeTranslator/ArrayAttributeTranslator.php
+++ b/src/AttributeTranslator/ArrayAttributeTranslator.php
@@ -11,12 +11,12 @@ use Yiisoft\Validator\AttributeTranslatorInterface;
  */
 final class ArrayAttributeTranslator implements AttributeTranslatorInterface
 {
+    /**
+     * @param array $translations Translations array where each key is an attribute name and the corresponding value is
+     * a translation.
+     * @psalm-param array<string,string> $translations
+     */
     public function __construct(
-        /**
-         * @var array Translations array where each key is an attribute name and the corresponding value is
-         * a translation.
-         * @psalm-var array<string,string>
-         */
         private array $translations,
     ) {
     }
--- a/src/AttributeTranslator/ArrayAttributeTranslator.php
+++ b/src/AttributeTranslator/ArrayAttributeTranslator.php
@@ -11,13 +11,11 @@ use Yiisoft\Validator\AttributeTranslatorInterface;
  */
 final class ArrayAttributeTranslator implements AttributeTranslatorInterface
 {
-    /**
-     * @param array $translations Translations array where each key is an attribute name and the corresponding name is
-     * a translation.
-     */
     public function __construct(
         /**
-         * @var array<string,string>
+         * @var array Translations array where each key is an attribute name and the corresponding value is
+         * a translation.
+         * @psalm-var array<string,string>
          */
         private array $translations,
     ) {
--- a/src/AttributeTranslator/NullAttributeTranslator.php
+++ b/src/AttributeTranslator/NullAttributeTranslator.php
@@ -7,7 +7,7 @@ namespace Yiisoft\Validator\AttributeTranslator;
 use Yiisoft\Validator\AttributeTranslatorInterface;
 
 /**
- * An attribute translator that returns attribute name as is.
+ * An attribute translator that returns attribute name as is without actually translating it.
  */
 final class NullAttributeTranslator implements AttributeTranslatorInterface
 {
--- a/src/AttributeTranslatorInterface.php
+++ b/src/AttributeTranslatorInterface.php
@@ -5,11 +5,13 @@ declare(strict_types=1);
 namespace Yiisoft\Validator;
 
 /**
- * An interface for translations of attributes names.
+ * Allows translating attribute names.
  */
 interface AttributeTranslatorInterface
 {
     /**
+     * Translate attribute name.
+     *
      * @param string $attribute The attribute name.
      *
      * @return string Translated attribute name.
--- a/src/AttributeTranslatorProviderInterface.php
+++ b/src/AttributeTranslatorProviderInterface.php
@@ -5,9 +5,15 @@ declare(strict_types=1);
 namespace Yiisoft\Validator;
 
 /**
- * Provides implementation of {@see AttributeTranslatorInterface} for translations of attribute names.
+ * Provides optional implementation of {@see AttributeTranslatorInterface} for translation of attribute names.
  */
 interface AttributeTranslatorProviderInterface
 {
+    /**
+     * @return AttributeTranslatorInterface|null Attribute translator instance to use. If null is returned,
+     * the validator uses default value configured.
+     *
+     * @see Validator constructor, `$defaultAttributeTranslator` argument.
+     */
     public function getAttributeTranslator(): ?AttributeTranslatorInterface;
 }
--- a/src/AttributeTranslatorProviderInterface.php
+++ b/src/AttributeTranslatorProviderInterface.php
@@ -10,10 +10,12 @@ namespace Yiisoft\Validator;
 interface AttributeTranslatorProviderInterface
 {
     /**
+     * Get attribute translator instance to use.
+     *
      * @return AttributeTranslatorInterface|null Attribute translator instance to use. If null is returned,
      * the validator uses default value configured.
      *
-     * @see Validator constructor, `$defaultAttributeTranslator` argument.
+     * @see Validator::$defaultAttributeTranslator (can be configured via constructor).
      */
     public function getAttributeTranslator(): ?AttributeTranslatorInterface;
 }
--- a/src/ValidationContext.php
+++ b/src/ValidationContext.php
@@ -6,35 +6,49 @@ namespace Yiisoft\Validator;
 
 use RuntimeException;
 use Yiisoft\Arrays\ArrayHelper;
+use Yiisoft\Validator\Rule\Each;
+use Yiisoft\Validator\Rule\Nested;
 
 /**
  * Validation context that might be taken into account when performing validation.
+ * The context is passed to both direct calls of {@see ValidatorInterface::validate()} and when {@see Nested}
+ * or {@see Each} is used.
  *
  * @psalm-import-type RulesType from ValidatorInterface
  */
 final class ValidationContext
 {
+    /**
+     * @var ValidatorInterface|null A validator instance. Null means context data was not set yet
+     * with {@see setContextDataOnce()}.
+     */
     private ?ValidatorInterface $validator = null;
 
     /**
-     * @var mixed The raw validated data.
+     * @var mixed The raw validated data. Null means context data was not set yet with {@see setContextDataOnce()}.
      */
     private mixed $rawData = null;
 
     /**
-     * @var DataSetInterface|null Data set the attribute belongs to. Null if data set not set.
+     * @var DataSetInterface|null Data set the attribute belongs to.
+     * Null if data set was not set yet with {@see setDataSet()}.
      */
     private ?DataSetInterface $dataSet = null;
 
     /**
-     * @var string|null Validated attribute name. Null if a single value is validated.
+     * @var string|null Validated data set attribute name. Null if a single value is validated.
      */
     private ?string $attribute = null;
 
+    /**
+     * @var AttributeTranslatorInterface|null Default attribute translator to use if attribute translator is not set.
+     */
     private ?AttributeTranslatorInterface $defaultAttributeTranslator = null;
 
     /**
      * @param array $parameters Arbitrary parameters.
+     * @param AttributeTranslatorInterface|null $attributeTranslator Optional attribute translator instance to use.
+     * If null is provided, or it's not specified, default passed through {@see setContextDataOnce()} is used.
      */
     public function __construct(
         private array $parameters = [],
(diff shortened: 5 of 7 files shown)


src/ValidationContext.php
```diff
@@ -6,42 +6,66 @@
...
+     * If null is provided, or it's not specified, default passed through {@see setContextDataOnce()} is used.
      */
     public function __construct(
         private array $parameters = [],
         private ?AttributeTranslatorInterface $attributeTranslator = null,
     ) {
     }
 
+    /**
+     * Set context data if it is not set yet.
```
vjik, line 60: Add recommendation don't use this method in userland. The following methods needs in other validator implementation or may be in validator decorator:
- `setContextDataOnce()`,
- `setDataSet()`,
- `setAttribute()`.

vjik (changes requested):

src/ValidationContext.php
samdark: I've marked these as internal methods.

src/AttributeTranslator/ArrayAttributeTranslator.php
```diff
@@ -11,6 +11,10 @@
  */
 final class ArrayAttributeTranslator implements AttributeTranslatorInterface
 {
+    /**
+     * @param array $translations Translations array where each key is an attribute name and the corresponding name is
+     * a translation.
+     */
     public function __construct(
         /**
          * @var array<string,string>
```
arogachev, line 20: Since it's a constructor argument and a class attribute at the same time, having docs in "var" is enough (for consistency also).

src/AttributeTranslator/ArrayAttributeTranslator.php
```diff
@@ -11,6 +11,10 @@
  */
 final class ArrayAttributeTranslator implements AttributeTranslatorInterface
 {
+    /**
+     * @param array $translations Translations array where each key is an attribute name and the corresponding name is
```
arogachev, line 15: Type `array` is less specific + it's an extra copy. Suggest to move it to "var".

src/AttributeTranslator/ArrayAttributeTranslator.php
arogachev, line 20: Method `translate()` is left without PHPDoc at all. Using "Ctrl + Q" shows that while some docs are inherited from interface (method description), the others are not (return description). Suggest to add even the copied version or discuss and possibly remove in other places later for consistency.

src/AttributeTranslator/NullAttributeTranslator.php
```diff
@@ -7,7 +7,7 @@
 use Yiisoft\Validator\AttributeTranslatorInterface;
 
 /**
- * An attribute translator that returns attribute name as is.
+ * An attribute translator that returns attribute name as is without actually translating it.
  */
 final class NullAttributeTranslator implements AttributeTranslatorInterface
 {
```
arogachev, line 13: Method `translate()` is left without PHPDoc at all (see https://github.com/yiisoft/validator/pull/473/files#r1052958155 for details).

src/AttributeTranslatorProviderInterface.php
```diff
@@ -5,9 +5,15 @@
...
 
 /**
- * Provides implementation of {@see AttributeTranslatorInterface} for translations of attribute names.
+ * Provides optional implementation of {@see AttributeTranslatorInterface} for translation of attribute names.
  */
 interface AttributeTranslatorProviderInterface
 {
+    /**
+     * @return AttributeTranslatorInterface|null Attribute translator instance to use. If null is returned,
+     * the validator uses default value configured.
```
arogachev, line 14:
```suggestion
     * @return AttributeTranslatorInterface|null Attribute translator instance to use. If null is returned, the
     * validator uses default value configured.
```

src/AttributeTranslatorProviderInterface.php
```diff
@@ -5,9 +5,15 @@
...
+     *
+     * @see Validator constructor, `$defaultAttributeTranslator` argument.
```
arogachev, line 16:
```suggestion
     * @see Validator::$defaultAttributeTranslator` (can be configured via constructor).
```

src/AttributeTranslatorProviderInterface.php
```diff
@@ -5,9 +5,15 @@
 namespace Yiisoft\Validator;
 
 /**
- * Provides implementation of {@see AttributeTranslatorInterface} for translations of attribute names.
+ * Provides optional implementation of {@see AttributeTranslatorInterface} for translation of attribute names.
  */
 interface AttributeTranslatorProviderInterface
 {
+    /**
```
arogachev, line 12: There is no method description.

src/AttributeTranslatorInterface.php
```diff
@@ -5,11 +5,13 @@
 namespace Yiisoft\Validator;
```
arogachev, line 5: A `TranslatorAttributeTranslator` marked as completed in #358, however constructor and method `translate()` are not documented.

src/ValidationContext.php
```diff
@@ -6,42 +6,68 @@
 
 use RuntimeException;
 use Yiisoft\Arrays\ArrayHelper;
+use Yiisoft\Validator\Rule\Each;
+use Yiisoft\Validator\Rule\Nested;
 
 /**
  * Validation context that might be taken into account when performing validation.
+ * The context is passed to both direct calls of {@see ValidatorInterface::validate()} and when {@see Nested}
```
arogachev, line 14: Not clear with `Nested` / `Each`.

src/ValidationContext.php
```diff
@@ -6,42 +6,68 @@
...
+ * or {@see Each} is used.
  *
  * @psalm-import-type RulesType from ValidatorInterface
  */
 final class ValidationContext
 {
+    /**
+     * @var ValidatorInterface|null A validator instance. Null means context data was not set yet
+     * with {@see setContextDataOnce()}.
```
arogachev, line 23:
```suggestion
     * @var ValidatorInterface|null A validator instance. `null` means context data was not set with
     * with {@see setContextDataOnce() yet.
```

src/ValidationContext.php
```diff
@@ -6,42 +6,68 @@
...
+     */
     private ?ValidatorInterface $validator = null;
 
     /**
-     * @var mixed The raw validated data.
+     * @var mixed The raw validated data. Null means context data was not set yet with {@see setContextDataOnce()}.
```
arogachev, line 28:
```suggestion
     * @var mixed The raw validated data. `null` means context data was not set with {@see setContextDataOnce()} yet.
```

src/ValidationContext.php
```diff
@@ -6,42 +6,68 @@
...
      */
     private mixed $rawData = null;
 
     /**
-     * @var DataSetInterface|null Data set the attribute belongs to. Null if data set not set.
+     * @var DataSetInterface|null Data set the attribute belongs to.
+     * Null if data set was not set yet with {@see setDataSet()}.
```
arogachev, line 34:
```suggestion
     * `null` if data set was not set with {@see setDataSet()} yet.
```

src/ValidationContext.php
```diff
@@ -6,42 +6,68 @@
...
      */
     private ?DataSetInterface $dataSet = null;
 
     /**
-     * @var string|null Validated attribute name. Null if a single value is validated.
+     * @var string|null Validated data set attribute name. Null if a single value is validated.
```
arogachev, line 39:
```suggestion
     * @var string|null Validated data set's attribute name. Null if a single value is validated.
```

src/ValidationContext.php
```diff
@@ -6,42 +6,68 @@
...
      */
     private ?string $attribute = null;
 
+    /**
+     * @var AttributeTranslatorInterface|null Default attribute translator to use if attribute translator is not set.
```
arogachev, line 44:
```suggestion
     * @var AttributeTranslatorInterface|null Default attribute translator to use if it's not set.
```

src/ValidationContext.php
```diff
@@ -6,42 +6,68 @@
...
+     */
     private ?AttributeTranslatorInterface $defaultAttributeTranslator = null;
 
     /**
      * @param array $parameters Arbitrary parameters.
+     * @param AttributeTranslatorInterface|null $attributeTranslator Optional attribute translator instance to use.
+     * If null is provided, or it's not specified, default passed through {@see setContextDataOnce()} is used.
```
arogachev, line 51:
```suggestion
     * If `null` is provided or it's not specified, a default one passed through {@see setContextDataOnce()} is used.
```

src/ValidationContext.php
```diff
@@ -6,42 +6,68 @@
...
      */
     private ?string $attribute = null;
 
+    /**
+     * @var AttributeTranslatorInterface|null Default attribute translator to use if attribute translator is not set.
+     */
     private ?AttributeTranslatorInterface $defaultAttributeTranslator = null;
 
     /**
      * @param array $parameters Arbitrary parameters.
```
arogachev, line 49: Suggest to use "var" and place right above corresponding arguments. Because they are also class attributes and for consistency.

src/ValidationContext.php
```diff
@@ -6,42 +6,68 @@
...
+     * Set context data if it is not set yet.
+     *
+     * @param ValidatorInterface $validator A validator instance.
+     * @param AttributeTranslatorInterface $attributeTranslator Attribute translator to use by default. If translator
+     * is specified via {@see setAttributeTranslator()}, it will be used instead.
+     * @param mixed $rawData The raw validated data.
+     *
+     * @internal
+     *
+     * @return $this
```
arogachev, line 69:
```suggestion
     * @return $this The same instance of validation context.
```

src/ValidationContext.php
```diff
@@ -58,6 +84,14 @@ public function setContextDataOnce(
         return $this;
     }
 
+    /**
+     * Set attribute translator to use.
+     *
+     * @param AttributeTranslatorInterface|null $attributeTranslator Attribute translator to use. If null,
```
arogachev, line 90:
```suggestion
     * @param AttributeTranslatorInterface|null $attributeTranslator Attribute translator to use. If `null`,
```

src/ValidationContext.php
```diff
@@ -58,6 +84,14 @@ public function setContextDataOnce(
...
+     * translator passed in {@see setContextData()} will be used.
+     *
+     * @return $this
```
arogachev, line 93:
```suggestion
     * @return $this The same instance of validation context.
```

src/ValidationContext.php
```diff
@@ -112,6 +146,8 @@ public function getDataSet(): DataSetInterface
```
arogachev, line 146: `validate()` method - return PHPDoc is missing.

src/ValidationContext.php
```diff
@@ -127,6 +163,11 @@ public function getAttribute(): ?string
         return $this->attribute;
     }
 
+    /**
+     * Get translated attribute name.
+     *
+     * @return string|null Translated attribute name.
```
arogachev, line 169: `null` case is not described.

src/ValidationContext.php
```diff
@@ -156,24 +201,35 @@ public function setAttribute(?string $attribute): self
...
     }
 
-    public function setParameter(string $key, mixed $value): self
+    /**
+     * Set parameter value.
+     *
+     * @param string $name Parameter name.
+     * @param mixed $value Parameter value.
+     *
+     * @return $this
```
arogachev, line 222:
```suggestion
     * @return $this The same instance of validation context.
```

src/ValidationContext.php
```diff
@@ -156,24 +201,35 @@ public function setAttribute(?string $attribute): self
...
+     */
+    public function setParameter(string $name, mixed $value): self
     {
-        $this->parameters[$key] = $value;
+        $this->parameters[$name] = $value;
         return $this;
     }
 
+    /**
+     * @return bool If attribute is missing.
```
arogachev, line 231:
```suggestion
     * @return bool If attribute is missing in a {@see $dataSet}.
```

arogachev (changes requested):
See suggestions.

src/AttributeTranslator/ArrayAttributeTranslator.php
samdark: It is IDE bug we need to fill in the issue tracker.

samdark: https://youtrack.jetbrains.com/issue/WI-70430/phpdoc-is-not-inherited-from-the-interface

src/ValidationContext.php
samdark: No, it's incorrect like that.

src/AttributeTranslator/NullAttributeTranslator.php
samdark: That is alright. It should be inherited from an interface but isn't due to IDE bug.

src/AttributeTranslator/ArrayAttributeTranslator.php
```diff
@@ -13,7 +13,9 @@ final class ArrayAttributeTranslator implements AttributeTranslatorInterface
 {
     public function __construct(
         /**
-         * @var array<string,string>
+         * @var array Translations array where each key is an attribute name and the corresponding value is
+         * a translation.
+         * @psalm-var array<string,string>
```
arogachev, line 18:
```suggestion
         * @var var array<string,string> Translations array where each key is an attribute name and the corresponding value is
         * a translation.
```

src/ValidationContext.php
```diff
@@ -14,34 +14,59 @@
...
+    private ?AttributeTranslatorInterface $defaultAttributeTranslator = null;
+
     public function __construct(
+        /**
+         * @var array Arbitrary parameters.
+         */
         private array $parameters = [],
+        /**
+         * @var AttributeTranslatorInterface|null $attributeTranslator Optional attribute translator instance to use.
+         * If `null` is provided, or it's not specified, a default translator passed through
```
arogachev, line 51:
```suggestion
         * If `null` is provided or it's not specified, a default translator passed through
```

arogachev (changes requested):
There are few more nitpicks. Other than that looks good. :+1:

src/AttributeTranslator/ArrayAttributeTranslator.php
samdark: There's no such type in phpdoc.

arogachev (approved):

PR #474, opened 2022-12-19
Clean up/add docs for Url and UrlHandler

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #358


src/Rule/Url.php
```diff
@@ -80,29 +93,45 @@ public function getName(): string
...
+     * fail. Note that in order to use IDN validation you have to install and enable `intl` PHP
+     * extension, otherwise an exception would be thrown.
+     */
+    public function isIDNEnabled(): bool
     {
         return $this->enableIDN;
     }
 
+    /**
+     * @return string A message used when the input it incorrect.
```
vjik, line 125:
```suggestion
     * @return string A message used when the input is incorrect.
```

src/Rule/Url.php
```diff
@@ -80,29 +93,45 @@ public function getName(): string
...
     }
 
-    public function isEnableIDN(): bool
+    /**
+     * @return bool Whether validation process should take into account IDN (internationalized
+     * domain names). `false` meaning that validation of URLs containing IDN will always
+     * fail. Note that in order to use IDN validation you have to install and enable `intl` PHP
+     * extension, otherwise an exception would be thrown.
+     */
+    public function isIDNEnabled(): bool
```
vjik, line 119:
```suggestion
    public function isIdnEnabled(): bool
```

src/Rule/Url.php
```diff
@@ -34,36 +34,49 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
         private array $validSchemes = ['http', 'https'],
         /**
-         * @var bool whether validation process should take into account IDN (internationalized
-         * domain names). Defaults to false meaning that validation of URLs containing IDN will always
+         * @var bool Whether validation process should take into account IDN (internationalized
+         * domain names). Defaults to `false` meaning that validation of URLs containing IDN will always
          * fail. Note that in order to use IDN validation you have to install and enable `intl` PHP
          * extension, otherwise an exception would be thrown.
          */
         private bool $enableIDN = false,
```
vjik, line 56:
```suggestion
        private bool $enableIdn = false,
```

src/Rule/Url.php
```diff
@@ -34,36 +34,49 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
+        /**
+         * @var string A message used when the input it incorrect.
+         */
         private string $incorrectInputMessage = 'The value must have a string type.',
+        /**
+         * @var string A message used when the value is not valid.
```
vjik, line 62: Add message parameters that may used in it (`{attribute}`, `{value}`).

src/Rule/Url.php
```diff
@@ -34,36 +34,49 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
-         * @var bool whether validation process should take into account IDN (internationalized
-         * domain names). Defaults to false meaning that validation of URLs containing IDN will always
+         * @var bool Whether validation process should take into account IDN (internationalized
+         * domain names). Defaults to `false` meaning that validation of URLs containing IDN will always
          * fail. Note that in order to use IDN validation you have to install and enable `intl` PHP
          * extension, otherwise an exception would be thrown.
          */
         private bool $enableIDN = false,
+        /**
+         * @var string A message used when the input it incorrect.
```
vjik, line 58: Add message parameters that may used in it (`{attribute}`, `{type}`).

src/Rule/Url.php
samdark: Why? The property is caps-ed as well as original method.

src/Rule/Url.php
samdark: Solved by adding `@see` reference.

src/Rule/Url.php
vjik: Usually in camelCase style acronyms used in lowercase also. For example, `URL` used as `url` almost always.

samdark: We need that in the code style docs, I guess. And fixed everywhere.

src/Rule/UrlHandler.php
```diff
@@ -36,9 +36,9 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
             return $result;
```
arogachev, line 36: The class description in the current state is a copy paste from corresponding rule. Needs to be adjusted. At least "this rule" part is incorrect and handler can be linked with rule.

src/Rule/UrlHandler.php
```diff
@@ -55,13 +55,28 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
         return $result;
     }
 
+    /**
```
arogachev, line 58: I'd mention here that it's just a wrapper over `idn_to_ascii` built-in function with a different handling of return value. A `{@link}` / `{@see}` tag to original function won't be redundant too in my opinion.

src/Rule/Url.php
```diff
@@ -34,36 +34,59 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
          * extension, otherwise an exception would be thrown.
          */
         private bool $enableIDN = false,
+        /**
+         * @var string A message used when the input is incorrect.
+         *
+         * You may use the following placeholders in the message:
+         *
+         * - `{attribute}`: the label of the attribute being validated.
+         * - `{value}`: the value of the attribute being validated.
```
arogachev, line 63:
```suggestion
         * - `{type}`: the value type.
```

src/Rule/Url.php
```diff
@@ -34,36 +34,59 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
-
         /**
-         * @var bool|callable|null
+         * @var bool|callable|null Whether to skip this rule if the value validated is empty.
+         *
+         * @see SkipOnEmptyInterface
          */
         private $skipOnEmpty = null,
+        /**
+         * @var bool Whether to skip this rule if any of the previous rules gave an error.
```
arogachev, line 82:
```suggestion
         * @var bool Whether to skip this rule if any of the previous rules gave an error.
         *
         * @see SkipOnErrorInterface
```

src/Rule/Url.php
```diff
@@ -34,36 +34,59 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
+         * You may use the following placeholders in the message:
+         *
+         * - `{attribute}`: the label of the attribute being validated.
+         * - `{value}`: the value of the attribute being validated.
+         */
         private string $message = 'This value is not a valid URL.',
-
         /**
-         * @var bool|callable|null
+         * @var bool|callable|null Whether to skip this rule if the value validated is empty.
```
arogachev, line 76:
```suggestion
         * @var bool|callable|null Whether to skip this rule if the validated value is empty.
```

src/Rule/Url.php
```diff
@@ -34,36 +34,59 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
          *
-         * Note that if you want to reuse the pattern in HTML5 input it should have ^ and $, should not have any
-         * modifiers and should not be case-insensitive.
+         * Note that if you want to reuse the pattern in HTML5 input, it should have `^` and `$`, should not have any
+         * modifiers, and should not be case-insensitive.
          */
         private string $pattern = '/^{schemes}:\/\/(([a-zA-Z0-9][a-zA-Z0-9_-]*)(\.[a-zA-Z0-9][a-zA-Z0-9_-]*)+)(?::\d{1,5})?([?\/#].*$|$)/',
         /**
-         * @var string[] list of URI schemes which should be considered valid. By default, http and https
+         * @var string[] List of URI schemes which should be considered valid. By default, http and https
```
arogachev, line 46:
```suggestion
         * @var string[] List of URI schemes which should be considered valid. By default, `'http'` and `'https'`
```

src/Rule/Url.php
```diff
@@ -34,36 +34,59 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
          * are considered to be valid schemes.
          */
         private array $validSchemes = ['http', 'https'],
         /**
-         * @var bool whether validation process should take into account IDN (internationalized
-         * domain names). Defaults to false meaning that validation of URLs containing IDN will always
+         * @var bool Whether validation process should take into account IDN (internationalized
```
arogachev, line 51:
```suggestion
         * @var bool Whether validation process must take IDN into account (internationalized
```

src/Rule/Url.php
```diff
@@ -34,36 +34,59 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
+         * domain names). Defaults to `false` meaning that validation of URLs containing IDN will always
```
arogachev, line 52:
```suggestion
         * domain names). Defaults to `false` means that validation of URLs containing IDN will always
```

src/Rule/Url.php
```diff
@@ -80,29 +103,53 @@ public function getName(): string
         return 'url';
     }
 
+    /**
```
arogachev, line 106: The method's description is missing. It's also actual for other methods.

src/Rule/Url.php
```diff
@@ -34,36 +34,59 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
          */
         private array $validSchemes = ['http', 'https'],
         /**
-         * @var bool whether validation process should take into account IDN (internationalized
-         * domain names). Defaults to false meaning that validation of URLs containing IDN will always
+         * @var bool Whether validation process should take into account IDN (internationalized
+         * domain names). Defaults to `false` meaning that validation of URLs containing IDN will always
          * fail. Note that in order to use IDN validation you have to install and enable `intl` PHP
          * extension, otherwise an exception would be thrown.
          */
```
arogachev, line 55: Suggest to add a link to a more detailed explanation.

src/Rule/Url.php
```diff
@@ -80,29 +103,53 @@ public function getName(): string
         return 'url';
     }
 
+    /**
+     * @return string Ready to use regular expression pattern used for URL validation.
```
arogachev, line 107:
```suggestion
     * @return string Ready to use regular expression pattern applied for URL validation.
```

arogachev: To avoid repetition `use` / `used`.

src/Rule/Url.php
```diff
@@ -80,29 +103,53 @@ public function getName(): string
...
      */
     public function getValidSchemes(): array
     {
         return $this->validSchemes;
     }
 
-    public function isEnableIDN(): bool
+    /**
+     * @return bool Whether validation process should take into account IDN (internationalized
+     * domain names). `false` meaning that validation of URLs containing IDN will always
```
arogachev, line 127:
```suggestion
     * @return bool Whether validation process should take IDN (internationalized domain names) into account).
     * . `false` meaning that validation of URLs containing IDN will always
```

src/Rule/Url.php
```diff
@@ -80,29 +103,53 @@ public function getName(): string
...
+     * fail. Note that in order to use IDN validation you have to install and enable `intl` PHP
+     * extension, otherwise an exception would be thrown.
```
arogachev, line 129:
```suggestion
     * extension, otherwise an exception will be thrown.
```

src/Rule/Url.php
```diff
@@ -34,36 +34,59 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
          * are considered to be valid schemes.
          */
         private array $validSchemes = ['http', 'https'],
         /**
-         * @var bool whether validation process should take into account IDN (internationalized
-         * domain names). Defaults to false meaning that validation of URLs containing IDN will always
+         * @var bool Whether validation process should take into account IDN (internationalized
+         * domain names). Defaults to `false` meaning that validation of URLs containing IDN will always
          * fail. Note that in order to use IDN validation you have to install and enable `intl` PHP
          * extension, otherwise an exception would be thrown.
```
arogachev, line 54:
```suggestion
         * extension, otherwise an exception will be thrown.
```

src/Rule/Url.php
arogachev: Or "being validated" / "that is validated".

src/Rule/Url.php
arogachev: For example - https://en.wikipedia.org/wiki/Internationalized_domain_name.

arogachev (changes requested):
See suggestions.

src/Rule/Url.php
```diff
@@ -20,10 +20,12 @@
 /**
  * Validates that the value is a valid HTTP or HTTPS URL.
```
arogachev, line 21: In my opinion the current description is a bit confusing, because the rule does not perform the actual validation / checking - it's done by handler.

src/Rule/Url.php
arogachev: This is actually wrong. in the other place it was "`false` meaning" -> "`false` means". This is OK because of "Defaults to" sentence start. I'll revert this.

arogachev: :heavy_check_mark:

arogachev (changes requested):
Let's fix the rule class description and it's ready to merge.

src/Rule/Url.php
```diff
@@ -32,42 +34,48 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
-         */
-        private bool $enableIDN = false,
+        private bool $enableIdn = false,
         private string $incorrectInputMessage = 'The value must have a string type.',
         private string $message = 'This value is not a valid URL.',
-
-        /**
-         * @var bool|callable|null
-         */
         private $skipOnEmpty = null,
```
arogachev, line 74:
```suggestion
        private mixed $skipOnEmpty = null,
```

src/Rule/Url.php
```diff
@@ -32,42 +34,48 @@ final class Url implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenI
...
-        /**
-         * @var WhenType
-         */
         private Closure|null $when = null,
     ) {
-        if ($enableIDN && !function_exists('idn_to_ascii')) {
+        if ($enableIdn && !function_exists('idn_to_ascii')) {
             // Tested via separate CI configuration (see ".github/workflows/build.yml").
             // @codeCoverageIgnoreStart
             throw new RuntimeException('In order to use IDN validation intl extension must be installed and enabled.');
```
arogachev, line 81: Maybe add "throws" tag?

src/Rule/Url.php
```diff
@@ -18,12 +18,14 @@
 use function function_exists;
 
 /**
- * Validates that the value is a valid HTTP or HTTPS URL.
+ * Defines validation rules for a value that is a valid HTTP or HTTPS URL.
```
arogachev, line 21:
```suggestion
 * Defines validation options for a value that is a valid HTTP or HTTPS URL.
```

arogachev (changes requested):
:+1:

arogachev (approved):

PR #475, opened 2022-12-19
Clean up/add docs for Ip and IpHandler

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #358

### DIFF
--- a/src/Rule/Ip.php
+++ b/src/Rule/Ip.php
@@ -104,6 +104,11 @@ final class Ip implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenIn
         private bool $allowNegation = false,
         /**
          * @var string A message used when the input it incorrect.
+         *
+         * You may use the following placeholders in the message:
+         *
+         * - `{attribute}`: the label of the attribute being validated.
+         * - `{value}`: the value of the attribute being validated.
          */
         private string $incorrectInputMessage = 'The value must have a string type.',
         /**
--- a/src/Rule/Ip.php
+++ b/src/Rule/Ip.php
@@ -25,6 +25,7 @@ use function strlen;
  * It also may change the value if normalization of IPv6 expansion is enabled.
  *
  * @psalm-import-type WhenType from WhenInterface
+ * @see IpHandler
  */
 #[Attribute(Attribute::TARGET_PROPERTY | Attribute::IS_REPEATABLE)]
 final class Ip implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenInterface, SkipOnEmptyInterface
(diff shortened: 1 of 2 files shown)


src/Rule/Ip.php
```diff
@@ -34,14 +34,16 @@ final class Ip implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenIn
...
     /**
-     * Negation char.
+     * Negation character.
      *
-     * Used to negate {@see $ranges} or {@see $network} or to negate validating value when {@see $allowNegation}
+     * Used to negate {@see $ranges} or {@see $network} or to negate value validated when {@see $allowNegation}
      * is used.
      */
-    private const NEGATION_CHAR = '!';
+    private const NEGATION_CHARACTER = '!';
```
vjik, line 42:
```suggestion
    public const NEGATION_CHARACTER = '!';
```

src/Rule/IpHandler.php
```diff
@@ -21,12 +21,12 @@
...
     /**
-     * Negation char.
+     * Negation character.
      *
-     * Used to negate {@see $ranges} or {@see $network} or to negate validating value when {@see $allowNegation}
+     * Used to negate {@see $ranges} or {@see $network} or to negate value validated when {@see $allowNegation}
      * is used.
      */
-    private const NEGATION_CHAR = '!';
+    private const NEGATION_CHARACTER = '!';
```
vjik, line 29:
```suggestion
```

Suggest use `Ip::NEGATION_CHARACTER`

src/Rule/Ip.php
samdark: Why making it public?

samdark: I think we should allow duplication in the internal implementation in this case to lower the number of useless things available for end user.

vjik (approved):

src/Rule/IpHandler.php
```diff
@@ -113,11 +109,11 @@ private static function validateVersion(
...
     ): Result|null {
-        if ($ipVersion === IpHelper::IPV6 && !$rule->isAllowIpv6()) {
+        if ($ipVersion === IpHelper::IPV6 && !$rule->isIpv6Allowed()) {
             return self::getGenericErrorResult($rule->getIpv6NotAllowedMessage(), $context, $value);
         }
 
-        if ($ipVersion === IpHelper::IPV4 && !$rule->isAllowIpv4()) {
+        if ($ipVersion === IpHelper::IPV4 && !$rule->isIpv4Allowed()) {
             return self::getGenericErrorResult($rule->getIpv4NotAllowedMessage(), $context, $value);
         }
```
arogachev, line 119: The methods are not documented:

- [ ] `getIpParsePattern()` - has description only.
- [ ] `validateValueParts`
- [ ] `validateVersion`
- [ ] `validateCidr`
- [ ] `getGenericErrorResult`

src/Rule/Ip.php
```diff
@@ -25,6 +25,8 @@
  * It also may change the value if normalization of IPv6 expansion is enabled.
```
arogachev, line 25: It was removed in the handler, maybe remove it here as well?

src/Rule/Ip.php
```diff
@@ -34,14 +36,16 @@ final class Ip implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenIn
...
     /**
-     * Negation char.
+     * Negation character.
      *
-     * Used to negate {@see $ranges} or {@see $network} or to negate validating value when {@see $allowNegation}
+     * Used to negate {@see $ranges} or {@see $network} or to negate value validated when {@see $allowNegation}
      * is used.
      */
-    private const NEGATION_CHAR = '!';
+    private const NEGATION_CHARACTER = '!';
```
arogachev, line 44: Maybe make it publicly available and reuse it handler?

src/Rule/Ip.php
```diff
@@ -56,152 +60,137 @@ final class Ip implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenIn
...
+     *     'any' // allows any other IP addresses
+     * ]);
+     * ```
+     *
+     * In this example, access is allowed for all the IPv4 and IPv6 addresses excluding the `192.168.10.0/24`
+     * subnet. IPv4 address `192.168.10.128` is also allowed, because it is listed before the restriction.
+     * @param bool|callable|null $skipOnEmpty Whether to skip this rule if the value validated is empty.
+     * See {@see SkipOnEmptyInterface}.
+     * @param bool $skipOnError Whether to skip this rule if any of the previous rules gave an error.
+     * {@see SkipOnErrorInterface}.
```
arogachev, line 172:
```suggestion
     * See {@see SkipOnErrorInterface}.
```

arogachev (changes requested):
See suggestions

src/Rule/Ip.php
```diff
@@ -56,152 +60,137 @@ final class Ip implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenIn
...
-         * @var bool|callable|null
-         */
         private mixed $skipOnEmpty = null,
         private bool $skipOnError = false,
-        /**
-         * @var WhenType
-         */
         private Closure|null $when = null,
     ) {
         if (!$this->allowIpv4 && !$this->allowIpv6) {
```
arogachev, line 196: Maybe add "throws" tags?

src/Rule/Ip.php
arogachev: Let's discuss / handle it separately - #484.

PR #481, opened 2022-12-20
Fixes in `ValidationContext`

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | 

- Actualize exception message and improve naming for method requiring validator.
- Use data set via getter in `isAttributeMissing()` method.

### DIFF
--- a/src/ValidationContext.php
+++ b/src/ValidationContext.php
@@ -76,7 +76,7 @@ final class ValidationContext
      */
     public function validate(mixed $data, callable|iterable|object|string|null $rules = null): Result
     {
-        $this->checkValidatorAndRawData();
+        $this->requireValidator();
 
         $currentDataSet = $this->dataSet;
         $currentAttribute = $this->attribute;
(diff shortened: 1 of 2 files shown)


src/ValidationContext.php
```diff
@@ -176,16 +176,16 @@ public function setParameter(string $key, mixed $value): self
 
     public function isAttributeMissing(): bool
     {
-        return $this->attribute !== null && $this->dataSet !== null && !$this->dataSet->hasAttribute($this->attribute);
+        return $this->attribute !== null && !$this->getDataSet()->hasAttribute($this->attribute);
```
vjik, line 179: It's throw exception when not set data set. Is it purpose of change?

arogachev: Yes. Because a data set is always set here for every context:

https://github.com/yiisoft/validator/blob/d5bc19b6939bdcc7907287ec6c49947a906f2244/src/Validator.php#L85

The lack of it is a programmer error.

vjik (approved):

Issue #476, opened 2022-12-19
Fix psalm errors


vjik:
With Psalm 5.3 new errors have appears in the package:

```
ERROR: ArgumentTypeCoercion - src/Helper/ObjectParser.php:314:62 - Argument 1 of ReflectionClass::getProperties expects 0|1|2|3|4|5|6|7|16|17|18|19|20|21|22|23|128|129|130|131|132|133|134|135|144|145|146|147|148|149|150|151|null, but parent type int provided (see https://psalm.dev/193)
        foreach ($this->getReflectionSource()->getProperties($this->propertyVisibility) as $property) {


ERROR: ArgumentTypeCoercion - src/Helper/ObjectParser.php:314:62 - Argument 1 of ReflectionObject::getProperties expects 0|1|2|3|4|5|6|7|16|17|18|19|20|21|22|23|128|129|130|131|132|133|134|135|144|145|146|147|148|149|150|151|null, but parent type int provided (see https://psalm.dev/193)
        foreach ($this->getReflectionSource()->getProperties($this->propertyVisibility) as $property) {
```

PR #488, opened 2022-12-23
Fix psalm errors

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #476

### DIFF
--- a/src/Helper/ObjectParser.php
+++ b/src/Helper/ObjectParser.php
@@ -355,7 +355,7 @@ final class ObjectParser
     /**
      * @psalm-param array $source Raw rules containing additional metadata besides rule instances.
      *
-     * @return array<int, RuleInterface>|array<string, list<RuleInterface>> An array of rules readies to use for the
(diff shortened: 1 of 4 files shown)


src/Helper/ObjectParser.php
```diff
@@ -351,21 +353,24 @@ private function getReflectionSource(): ReflectionObject|ReflectionClass
     }
 
     /**
-     * @psalm-param RulesCache $source Raw rules containing additional metadata besides rule instances.
+     * @psalm-param array $source Raw rules containing additional metadata besides rule instances.
      *
-     * @return array<int, RuleInterface>|array<string, list<RuleInterface>> An array of rules ready to use for the
+     * @return array<int, RuleInterface>|array<string, list<RuleInterface>> An array of rules readies to use for the
```
samdark, line 358:
```suggestion
     * @return array<int, RuleInterface>|array<string, list<RuleInterface>> An array of rules ready to use for the
```

samdark (approved):

PR #489, opened 2022-12-24
More phpdoc

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | part of #358

### DIFF
--- a/src/Rule/AtLeast.php
+++ b/src/Rule/AtLeast.php
@@ -17,6 +17,8 @@ use Yiisoft\Validator\WhenInterface;
 /**
  * Defines validation options to check that a minimum number of specified attributes are filled.
  *
+ * Arrays and objects with public properties are supported as values validated.
+ *
  * @see AtLeastHandler
  *
  * @psalm-import-type WhenType from WhenInterface
@@ -29,19 +31,19 @@ final class AtLeast implements RuleWithOptionsInterface, SkipOnErrorInterface, W
     use WhenTrait;
 
     /**
-     * @param array $attributes string[] The list of required attributes that will be checked.
+     * @param string[] $attributes The list of required attributes that will be checked.
      * @param int $min The minimum required quantity of filled attributes to pass the validation. Defaults to 1.
      * @param string $incorrectInputMessage A message used when the input is incorrect.
      *
      * You may use the following placeholders in the message:
      *
-     * - `{attribute}`: the label of the attribute being validated.
-     * - `{type}`: the type of the attribute being validated.
+     * - `{attribute}`: the translated label of the attribute being validated.
+     * - `{type}`: the type of the value being validated.
      * @param string $message A message used when the value is not valid.
      *
      * You may use the following placeholders in the message:
      *
-     * - `{attribute}`: the label of the attribute being validated.
+     * - `{attribute}`: the translated label of the attribute being validated.
      * - `{min}`: the minimum number of attribute values that was not met.
      * @param bool|callable|null $skipOnEmpty Whether to skip this rule if the value validated is empty.
      * See {@see SkipOnEmptyInterface}.
--- a/src/Rule/AtLeast.php
+++ b/src/Rule/AtLeast.php
@@ -15,7 +15,9 @@ use Yiisoft\Validator\SkipOnErrorInterface;
 use Yiisoft\Validator\WhenInterface;
 
 /**
- * Checks if at least {@see AtLeast::$min} of specified attributes are filled.
+ * Defines validation options to check that a minimum number of specified attributes are filled.
+ *
+ * @see AtLeastHandler
  *
  * @psalm-import-type WhenType from WhenInterface
  */
@@ -26,28 +28,36 @@ final class AtLeast implements RuleWithOptionsInterface, SkipOnErrorInterface, W
     use SkipOnErrorTrait;
     use WhenTrait;
 
+    /**
+     * @param array $attributes string[] The list of required attributes that will be checked.
+     * @param int $min The minimum required quantity of filled attributes to pass the validation. Defaults to 1.
+     * @param string $incorrectInputMessage A message used when the input is incorrect.
+     *
+     * You may use the following placeholders in the message:
+     *
+     * - `{attribute}`: the label of the attribute being validated.
+     * - `{type}`: the type of the attribute being validated.
+     * @param string $message A message used when the value is not valid.
+     *
+     * You may use the following placeholders in the message:
+     *
+     * - `{attribute}`: the label of the attribute being validated.
+     * - `{min}`: the minimum number of attribute values that was not met.
+     * @param bool|callable|null $skipOnEmpty Whether to skip this rule if the value validated is empty.
+     * See {@see SkipOnEmptyInterface}.
+     * @param bool $skipOnError Whether to skip this rule if any of the previous rules gave an error.
+     * See {@see SkipOnErrorInterface}.
+     * @param Closure|null $when A callable to define a condition for applying the rule.
+     * See {@see WhenInterface}.
+     * @psalm-param WhenType $when
+     */
     public function __construct(
-        /**
-         * @var string[] The list of required attributes that will be checked.
-         */
         private array $attributes,
-        /**
-         * @var int The minimum required quantity of filled attributes to pass the validation. Defaults to 1.
-         */
         private int $min = 1,
         private string $incorrectInputMessage = 'Value must be an array or an object.',
-        /**
-         * @var string Message to display in case of error.
-         */
         private string $message = 'The data must have at least "{min}" filled attributes.',
-        /**
-         * @var bool|callable|null
-         */
         private mixed $skipOnEmpty = null,
         private bool $skipOnError = false,
-        /**
-         * @psalm-var WhenType
-         */
         private Closure|null $when = null
     ) {
     }
@@ -58,23 +68,48 @@ final class AtLeast implements RuleWithOptionsInterface, SkipOnErrorInterface, W
     }
 
     /**
-     * @return string[]
+     * Get the list of required attributes that will be checked.
+     *
+     * @return string[] The list of attributes.
+     *
+     * @see $attributes
      */
     public function getAttributes(): array
     {
         return $this->attributes;
     }
 
+    /**
+     * Get the minimum required quantity of filled attributes to pass the validation.
+     *
+     * @return int Minimum require quantity.
+     *
+     * @see $min
+     */
     public function getMin(): int
     {
         return $this->min;
     }
 
+    /**
+     * Get the message used when the input is incorrect.
+     *
+     * @return string Error message.
+     *
+     * @see $incorrectInputMessage
+     */
     public function getIncorrectInputMessage(): string
     {
         return $this->incorrectInputMessage;
     }
 
+    /**
+     * Get the message used when the value is not valid.
+     *
+     * @return string Error message.
+     *
+     * @see $message
+     */
     public function getMessage(): string
     {
         return $this->message;
--- a/src/Rule/Callback.php
+++ b/src/Rule/Callback.php
@@ -9,7 +9,6 @@ use Closure;
 use InvalidArgumentException;
 use ReflectionObject;
 use Yiisoft\Validator\AfterInitAttributeEventInterface;
-use Yiisoft\Validator\Result;
 use Yiisoft\Validator\Rule\Trait\SkipOnEmptyTrait;
 use Yiisoft\Validator\Rule\Trait\SkipOnErrorTrait;
 use Yiisoft\Validator\Rule\Trait\WhenTrait;
@@ -38,9 +37,9 @@ final class Callback implements
     use WhenTrait;
 
     /**
-     * @var object|null The object being validated.
+     * @var object|null The object being validated. Null if PHP attributes aren't used.
      */
-    private ?object $objectValidated = null;
+    private ?object $validatedObject = null;
 
     /**
      * @param callable|null $callback Callable with the `function ($value, $rule, $context): Result` signature that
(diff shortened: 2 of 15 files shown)


src/Rule/Equal.php
```diff
@@ -19,42 +19,62 @@
...
+     * @param string|null $targetAttribute The attribute to be equal to. When both this property and {@see $targetValue} are set, the
+     * {@see $targetValue} takes precedence.
+     * @param string $incorrectInputMessage A message used when the input is incorrect.
+     *
+     * You may use the following placeholders in the message:
+     *
+     * - `{attribute}`: the label of the attribute being validated.
+     * - `{type}`: the type of the attribute being validated.
+     * @param string $incorrectDataSetTypeMessage A message used when the attribute value returned from a custom
+     * data set s not scalar.
```
samdark, line 41:
```suggestion
     * data set is not a scalar.
```

src/Rule/AtLeast.php
```diff
@@ -15,7 +15,9 @@
 use Yiisoft\Validator\WhenInterface;
 
 /**
- * Checks if at least {@see AtLeast::$min} of specified attributes are filled.
+ * Defines validation options to check that a minimum number of specified attributes are filled.
```
vjik, line 18: Add info about supported values: array and object with public properties.

src/Rule/Count.php
```diff
@@ -108,11 +118,25 @@ public function getName(): string
         return 'count';
     }
 
+    /**
+     * Get error message used when the value is neither an array nor implementing {@see \Countable} interface.
```
vjik, line 122:
```suggestion
     * Get error message used when the value is neither an array nor implementing {@see Countable} interface.
```

vjik (changes requested):
Need fix psalm errors.

src/Rule/Compare.php
```diff
@@ -15,6 +15,20 @@
 use Yiisoft\Validator\WhenInterface;
 
 /**
+ * Abstract base for all the comparison validation options.
+ *
+ * The value being compared with {@see Compare::$targetValue} or {@see Compare::$targetAttribute}, which is set
+ * in the constructor.
+ *
+ * It supports different comparison operators, specified
```
arogachev, line 23: Suggest to list them with equivalent code. For example:

- {@see Equal} is a shortcut for `new CompareTo(operator: '===')`.

and so on.

Also mention `CompareTo`.

src/Rule/Compare.php
```diff
@@ -15,6 +15,20 @@
 use Yiisoft\Validator\WhenInterface;
 
 /**
+ * Abstract base for all the comparison validation options.
```
arogachev, line 18:
```suggestion
 * Abstract base for all the comparison validation rules.
```

arogachev: Or rule classes / classes.

src/Rule/Equal.php
```diff
@@ -10,7 +10,7 @@
 use Yiisoft\Validator\WhenInterface;
 
 /**
- * Validates if the specified value is equal to another value or attribute.
+ * Defines validation options to check that the specified value is equal to another value or attribute.
```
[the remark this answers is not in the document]
arogachev: The same for other comparison rule shortcuts.

src/Rule/Count.php
```diff
@@ -38,59 +40,67 @@ final class Count implements
...
     private ?object $objectValidated = null;
 
+    /**
+     * @param int|null $min Minimum number of items. null means no minimum number limit. Can't be combined with
+     * {@see $exactly}. See {@see $lessThanMinMessage} for the customized message for a value with too few items.
+     * @param int|null $max Maximum number of items. null means no maximum number limit. Can't be combined with
+     * {@see $exactly}. See {@see $greaterThanMaxMessage} for the customized message for a value with too many items.
+     * @param int|null $exactly Exact number of items. `null` means no strict comparison. Mutually exclusive with
+     * {@see $min} and {@see $max}.
+     * @param string $incorrectInputMessage Error message used when the value is neither an array nor implementing
```
arogachev, line 55:
```suggestion
     * @param string $incorrectInputMessage Error message used when the value is neither an array nor a class implementing
```

src/Rule/Count.php
arogachev, line 122:
```suggestion
     * Get error message used when the value is neither an array nor a class implementing {@see \Countable} interface.
```

src/Rule/AtLeast.php
arogachev: Agree. It also could be added in the corresponding property too.

src/Rule/AtLeast.php
```diff
@@ -15,7 +15,9 @@
 use Yiisoft\Validator\WhenInterface;
```
arogachev, line 15: Need to fix new Psalm errors:

```
ERROR: MixedReturnTypeCoercion - src/Rule/AtLeast.php:73:16 - The declared return type 'array<array-key, string>' for Yiisoft\Validator\Rule\AtLeast::getAttributes is more specific than the inferred return type 'array<array-key, mixed>' (see https://psalm.dev/197)
     * @return string[] The list of attributes.


ERROR: MixedReturnTypeCoercion - src/Rule/AtLeast.php:79:16 - The type 'array<array-key, mixed>' is more general than the declared return type 'array<array-key, string>' for Yiisoft\Validator\Rule\AtLeast::getAttributes (see https://psalm.dev/197)
        return $this->attributes;
```

src/Rule/Callback.php
```diff
@@ -9,6 +9,7 @@
 use InvalidArgumentException;
 use ReflectionObject;
 use Yiisoft\Validator\AfterInitAttributeEventInterface;
+use Yiisoft\Validator\Result;
```
arogachev, line 12: The import is unused. Need to delete.

src/Rule/Callback.php
```diff
@@ -18,6 +19,10 @@
 use Yiisoft\Validator\WhenInterface;
 
 /**
+ * Defines validation options to validating the value using a callback.
```
arogachev, line 22: Maybe add a little bit more explanation? For example that it can be used as an alternative for creating custom rules with handlers - simpler, but not reusable, etc.

src/Rule/Callback.php
```diff
@@ -32,23 +37,29 @@ final class Callback implements
     use SkipOnErrorTrait;
     use WhenTrait;
 
+    /**
+     * @var object|null The object being validated.
+     */
     private ?object $objectValidated = null;
```
arogachev, line 43:
```suggestion
    private ?object $validatedObject = null;
```

arogachev: Otherwise it can be understood as boolean - "was the object validated?".

src/Rule/Callback.php
arogachev: Also the separate usage / usage for a class method can be mentioned too.

src/Rule/Callback.php
```diff
@@ -32,23 +37,29 @@ final class Callback implements
...
 
+    /**
+     * @param callable|null $callback Callable with the `function ($value, $rule, $context): Result` signature that
```
arogachev, line 46: Maybe add the note about `$callback` and `$method` about being mutually exclusive with each other?

src/Rule/Callback.php
```diff
@@ -32,23 +37,29 @@ final class Callback implements
...
+     * @param string|null $method Name of a validated object method with the `function ($value, $rule, $context): Result`
+     * signature that performs the validation.
+     * @param bool|callable|null $skipOnEmpty Whether to skip this rule if the value validated is empty.
+     * See {@see SkipOnEmptyInterface}.
+     * @param bool $skipOnError Whether to skip this rule if any of the previous rules gave an error.
+     * See {@see SkipOnErrorInterface}.
+     * @param Closure|null $when A callable to define a condition for applying the rule.
+     * See {@see WhenInterface}.
+     * @psalm-param WhenType $when
+     */
```
arogachev, line 57: Add `@throws` tags. Also actual for other places.

src/Rule/Callback.php
```diff
@@ -32,23 +37,29 @@ final class Callback implements
     use SkipOnErrorTrait;
     use WhenTrait;
 
+    /**
+     * @var object|null The object being validated.
```
arogachev, line 41: Add explanation of `null` - that it's set only when using PHP attributes.

src/Rule/Callback.php
arogachev: It's weird that it was not removed automatically. :thinking:

src/Rule/CompareTo.php
```diff
@@ -7,17 +7,18 @@
 use Attribute;
 
 /**
- * Validates the specified value with another value or attribute.
+ * Defines validation options to check that the specified value match with another value or attribute.
```
arogachev, line 10:
```suggestion
 * Defines validation options to check that the specified value matches with another value or attribute.
```

src/Rule/CompareTo.php
```diff
@@ -7,17 +7,18 @@
...
  *
  * The value being compared with a constant {@see CompareTo::$compareValue}, which is set
  * in the constructor.
  *
- * It supports different comparison operators, specified
- * via the {@see CompareTo::$operator}.
+ * It supports different comparison operators, specified via the {@see CompareTo::$operator}.
```
arogachev, line 15: Suggest to list them with equivalent code. For example:

```
- {@see GreaterThan} is a shortcut for `new CompareTo(operator: '>')`.
```

and so on.

arogachev (changes requested):
See suggestions.

src/Rule/Callback.php
samdark: I think that belongs to readme.

src/Rule/Callback.php
samdark: That's in the exception so the issue will be visible right away.

src/Rule/Compare.php
```diff
@@ -49,51 +66,66 @@ abstract class Compare implements RuleWithOptionsInterface, SkipOnEmptyInterface
...
+     *
+     * You may use the following placeholders in the message:
+     *
+     * - `{attribute}`: the label of the attribute being validated.
+     * - `{targetValue}`: the constant value to be compared with.
+     * - `{targetAttribute}`: the name of the attribute to be compared with.
+     * - `{targetValueOrAttribute}`: the constant value to be compared with or, if it's absent, the name of
+     *   the attribute to be compared with.
+     * - `{value}`: the value of the attribute being validated.
+     * @param string $type The type of the values being compared. Either {@see Compare::TYPE_STRING}
```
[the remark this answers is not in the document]
samdark: No. It's not an int mask.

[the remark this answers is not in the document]
arogachev: The type is different, but the mask still can be added.

src/Rule/Callback.php
arogachev: I think having it mentioned in the docs additionally won't be redundant, like in `LimitTrait` for example.

src/Rule/Callback.php
```diff
@@ -32,23 +36,32 @@ final class Callback implements
     use SkipOnErrorTrait;
     use WhenTrait;
 
-    private ?object $objectValidated = null;
-
+    /**
+     * @var object|null The object being validated. Null if PHP attributes aren't used.
```
arogachev, line 40:
```suggestion
     * @var object|null The object being validated. `null` if PHP attributes aren't used.
```

src/Rule/Callback.php
```diff
@@ -32,23 +36,32 @@ final class Callback implements
...
+     * signature that performs the validation.
+     * @param bool|callable|null $skipOnEmpty Whether to skip this rule if the value validated is empty.
+     * See {@see SkipOnEmptyInterface}.
+     * @param bool $skipOnError Whether to skip this rule if any of the previous rules gave an error.
+     * See {@see SkipOnErrorInterface}.
+     * @param Closure|null $when A callable to define a condition for applying the rule.
+     * See {@see WhenInterface}.
+     * @psalm-param WhenType $when
+     *
+     * @throws InvalidArgumentException When either no callback or method is specified or
```
arogachev, line 57:
```suggestion
     * @throws InvalidArgumentException When neither {@see $callback} nor {@see $method} is specified or
```

src/Rule/Count.php
```diff
@@ -108,11 +118,25 @@ public function getName(): string
         return 'count';
     }
 
+    /**
+     * Get error message used when the value is neither an array nor a class implementing {@see \Countable} interface.
```
arogachev, line 122:
```suggestion
     * Get error message used when the value is neither an array nor an object implementing {@see \Countable} interface.
```

src/Rule/Count.php
```diff
@@ -38,59 +40,67 @@ final class Count implements
...
     private ?object $objectValidated = null;
 
+    /**
+     * @param int|null $min Minimum number of items. null means no minimum number limit. Can't be combined with
+     * {@see $exactly}. See {@see $lessThanMinMessage} for the customized message for a value with too few items.
+     * @param int|null $max Maximum number of items. null means no maximum number limit. Can't be combined with
+     * {@see $exactly}. See {@see $greaterThanMaxMessage} for the customized message for a value with too many items.
+     * @param int|null $exactly Exact number of items. `null` means no strict comparison. Mutually exclusive with
+     * {@see $min} and {@see $max}.
+     * @param string $incorrectInputMessage Error message used when the value is neither an array nor a class
```
arogachev, line 55:
```suggestion
     * @param string $incorrectInputMessage Error message used when the value is neither an array nor an object
```

arogachev (changes requested):
There are 5 unresolved suggestions from the previous review and 4 new.

src/Rule/Compare.php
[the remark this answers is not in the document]
arogachev: I mean this - https://psalm.dev/docs/annotating_code/typing_in_psalm/#specifying-stringint-options-aka-enums.

src/Rule/Callback.php
arogachev: Closing because it will be no longer actual after merging #464.

arogachev (approved):

PR #490, opened 2022-12-25
More phpdoc

samdark:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | part of #358


src/Rule/HasLength.php
```diff
@@ -36,56 +38,64 @@ final class HasLength implements
     use SkipOnErrorTrait;
     use WhenTrait;
 
+    /**
+     * @param int|null $min Minimum length. null means no minimum length limit. Can't be combined with
```
vjik, line 42:
```suggestion
     * @param int|null $min Minimum length. `null` means no minimum length limit. Can't be combined with
```

src/Rule/HasLength.php
```diff
@@ -36,56 +38,64 @@ final class HasLength implements
...
+     * {@see $exactly}. See {@see $lessThanMinMessage} for the customized message for a too short string.
+     * @param int|null $max maximum length. null means no maximum length limit. Can't be combined with
```
vjik, line 44:
```suggestion
     * @param int|null $max maximum length. `null` means no maximum length limit. Can't be combined with
```

src/Rule/HasLength.php
```diff
@@ -17,10 +17,12 @@
 use Yiisoft\Validator\WhenInterface;
 
 /**
- * Validates that the value is of certain length.
+ * Defines validation options to check that the value is of a certain length.
```
arogachev, line 20:
```suggestion
 * Defines validation options to check that the string is of a certain length.
```

src/Rule/HasLength.php
```diff
@@ -17,10 +17,12 @@
...
  *
  * Note, this rule should only be used with strings.
```
arogachev, line 22:
```suggestion
```

src/Rule/HasLength.php
```diff
@@ -36,56 +38,64 @@ final class HasLength implements
...
+     * @param int|null $max maximum length. null means no maximum length limit. Can't be combined with
+     * {@see $exactly}. See {@see $greaterThanMaxMessage} for the customized message for a too long string.
+     * @param int|null $exactly Exact length. `null` means no strict comparison. Mutually exclusive with
+     * {@see $min} and {@see $max}.
+     * @param string $incorrectInputMessage Error message used when the value is not a string.
+     *
+     * You may use the following placeholders in the message:
+     *
+     * - `{attribute}`: the label of the attribute being validated.
+     * - `{type}`: the type of the attribute being validated.
```
arogachev, line 53:
```suggestion
     * - `{type}`: the type of the value being validated.
```

arogachev: Because there is a case with single value when attribute is not presented at all.

arogachev: Actual for other occurrences too.

src/Rule/HasLength.php
```diff
@@ -36,56 +38,64 @@ final class HasLength implements
...
+     * @param string $lessThanMinMessage Error message used when the length of the value is smaller than {@see $min}.
+     *
+     * You may use the following placeholders in the message:
+     *
+     * - `{attribute}`: the label of the attribute being validated.
```
arogachev, line 58:
```suggestion
     * - `{attribute}`: the translated label of the attribute being validated.
```

arogachev: Actual for other occurrences too.

src/Rule/HasLength.php
```diff
@@ -103,11 +113,26 @@ public function getName(): string
         return 'hasLength';
     }
 
+    /**
+     * Get error message used when the value is neither an array nor implementing {@see \Countable} interface.
```
arogachev, line 117:
```suggestion
     * Get error message used when the value is neither an array nor implementing {@see Countable} interface.
```

arogachev: And add import. Not critical.

src/Rule/HasLengthHandler.php
```diff
@@ -13,9 +13,9 @@
 use function is_string;
 
 /**
- * Validates that the value is of certain length.
+ * Validates that the value is of a certain length.
```
arogachev, line 16:
```suggestion
 * Validates that the value is a string of a certain length.
```

src/Rule/In.php
```diff
@@ -28,30 +30,33 @@ final class In implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenIn
     use SkipOnErrorTrait;
     use WhenTrait;
 
+    /**
+     * @param iterable $values A set of values to check against.
```
arogachev, line 34: Need to mention that subarrays are not supported and recommend to use `Subset` instead. Maybe in the class description too.

src/Rule/In.php
```diff
@@ -28,30 +30,33 @@ final class In implements RuleWithOptionsInterface, SkipOnErrorInterface, WhenIn
...
+     * @psalm-param iterable<scalar> $values
+     *
+     * @param bool $strict Whether the comparison is strict (both type and value must be the same).
```
arogachev, line 37: I'd complement this with `Boolean::$strict` info (more detailed explanation, links to PHP docs).

src/Rule/In.php
```diff
@@ -61,21 +66,42 @@ public function getName(): string
         return 'inRange';
     }
 
+    /**
+     * Get a set of values to check against.
+     *
+     * @return iterable A set of values.
```
arogachev, line 72:
```suggestion
     * @return iterable A set of scalar values.
```

arogachev, line 72: `psalm-return` can be added.

src/Rule/In.php
```diff
@@ -61,21 +66,42 @@ public function getName(): string
...
+     *
+     * @return bool Whether to invert the validation logic.
+     */
     public function isNot(): bool
     {
         return $this->not;
     }
 
+    /**
+     * Get error message when the value is not in a set of value.
```
arogachev, line 101:
```suggestion
     * Get error message when the value is not in a set of {@see $values}.
```

src/Rule/Number.php
```diff
@@ -15,12 +15,14 @@
 use Yiisoft\Validator\WhenInterface;
 
 /**
- * Validates that the value is a number.
+ * Defines validation options to check that the value is a number.
  *
  * The format of the number must match the regular expression specified in {@see Number::$integerPattern}
  * or {@see Number::$numberPattern}. Optionally, you may configure the {@see Number::min()} and {@see Number::max()}
```
arogachev, line 21: In case of such explanation, `$exactly` can be mentioned too.

src/Rule/Number.php
```diff
@@ -30,50 +32,55 @@ final class Number implements RuleWithOptionsInterface, SkipOnErrorInterface, Wh
     use SkipOnErrorTrait;
     use WhenTrait;
 
+    /**
+     * @param bool $integerOnly Whether the value can only be an integer. Defaults to false.
```
arogachev, line 36:
```suggestion
     * @param bool $integerOnly Whether the value can only be an integer. Defaults to `false`.
```

src/Rule/Number.php
```diff
@@ -30,50 +32,55 @@ final class Number implements RuleWithOptionsInterface, SkipOnErrorInterface, Wh
...
+     * @param float|int|null $min Lower limit of the number. Defaults to null, meaning no lower limit.
```
arogachev, line 37:
```suggestion
     * @param float|int|null $min Lower limit of the number. Defaults to `null`, meaning no lower limit.
```

src/Rule/Number.php
```diff
@@ -30,50 +32,55 @@ final class Number implements RuleWithOptionsInterface, SkipOnErrorInterface, Wh
...
+     * See {@see $tooSmallMessage} for the customized message used when the number is too small.
+     * @param float|int|null $max Upper limit of the number. Defaults to null, meaning no upper limit.
```
arogachev, line 39:
```suggestion
     * @param float|int|null $max Upper limit of the number. Defaults to `null`, meaning no upper limit.
```

src/Rule/Number.php
```diff
@@ -30,50 +32,55 @@ final class Number implements RuleWithOptionsInterface, SkipOnErrorInterface, Wh
...
+     * See {@see $tooBigMessage} for the customized message used when the number is too big.
+     * @param string $incorrectInputMessage Error message used when the value is neither an array nor implementing
+     * {@see \Countable} interface.
```
arogachev, line 42: This is wrong. Probably a leftover from `HasLength` copy paste.

src/Rule/Number.php
```diff
@@ -83,55 +90,117 @@ public function getName(): string
         return 'number';
     }
 
-    public function isAsInteger(): bool
+    /**
+     * Whether the value can only be an integer.
+     *
+     * @return bool Whether the value can only be an integer. Defaults to false.
```
arogachev, line 96:
```suggestion
     * @return bool Whether the value can only be an integer. Defaults to `false`.
```

src/Rule/Number.php
```diff
@@ -83,55 +90,117 @@ public function getName(): string
...
+     * @see $integerOnly
+     */
+    public function isIntegerOnly(): bool
     {
-        return $this->asInteger;
+        return $this->integerOnly;
     }
 
+    /**
+     * Get lower limit of the number. Null means no lower limit.
```
arogachev, line 106:
```suggestion
     * Get lower limit of the number. `null` means no lower limit.
```

src/Rule/Number.php
```diff
@@ -83,55 +90,117 @@ public function getName(): string
...
+     *
+     * @see $min
+     */
     public function getMin(): float|int|null
     {
         return $this->min;
     }
 
+    /**
+     * Get upper limit of the number. Null means no upper limit.
```
arogachev, line 118:
```suggestion
     * Get upper limit of the number. `null` means no upper limit.
```

src/Rule/Number.php
arogachev: Also maybe move `null` note to return's description?

src/Rule/Number.php
```diff
@@ -83,55 +90,117 @@ public function getName(): string
...
+     *
+     * @see $max
+     */
     public function getMax(): float|int|null
     {
         return $this->max;
     }
 
+    /**
+     * Get error message used when the value is neither an array nor implementing
```
arogachev, line 130: Need to remove copy paste.

src/Rule/NumberHandler.php
```diff
@@ -18,6 +18,8 @@
  * The format of the number must match the regular expression specified in {@see Number::$integerPattern}
```
arogachev, line 18: Such details can be removed in the handler, the link for corresponding rule is enough in my opinion (or a shortened version of description). But I prefer the link more.

src/Rule/RegexHandler.php
```diff
@@ -14,7 +14,7 @@
 /**
  * Validates that the value matches the pattern specified in constructor.
```
arogachev, line 15: I'd either mentioned possible inversion option, because current statement is not 100% correct or just provided the link to the handler.

arogachev (approved):

arogachev (changes requested):
See suggestions.

src/Rule/Number.php
samdark: There's no `$exactly`.

src/Rule/RegexHandler.php
samdark: The link is there.

samdark:
Fixed it. Please re-check.

src/Rule/Number.php
arogachev: Oops, sorry, this will be actual only after merging #422.

arogachev (approved):

src/Rule/Number.php
```diff
@@ -83,55 +89,116 @@ public function getName(): string
...
     {
-        return $this->asInteger ? 'Value must be an integer.' : 'Value must be a number.';
+        return $this->integerOnly ? 'Value must be an integer.' : 'Value must be a number.';
     }
 
     public function getOptions(): array
     {
         return [
-            'asInteger' => $this->asInteger,
+            'asInteger' => $this->integerOnly,
```
cebe, line 201: option should be renamed if property is renamed, right?

arogachev: Yes.

PR #505, opened 2023-01-09
Docs, part 16

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #358

### DIFF
--- a/src/Rule/Nested.php
+++ b/src/Rule/Nested.php
@@ -45,32 +45,63 @@ use function sprintf;
  * An example with blog post:
  *
  * ```php
- * $rules = [
- *     new Nested([
- *         'title' => [new HasLength(max: 255)],
- *          // One-to-one relation
- *         'author' => new Nested([
- *             'name' => [new HasLength(min: 1)],
+ * $rule = new Nested([
+ *     'title' => [new HasLength(max: 255)],
+ *     // One-to-one relation
+ *     'author' => new Nested([
+ *         'name' => [new HasLength(min: 1)],
+ *     ]),
+ *     // One-to-many relation
+ *     'files' => new Each([
+ *          new Nested([
+ *             'url' => [new Url()],
  *         ]),
- *         // One-to-many relation
- *         'files' => new Each([
- *              new Nested([
- *                 'url' => [new Url()],
- *             ]),
- *         ]),
- *     ]);
- * ];
+ *     ]),
+ * ]);
  * ```
  *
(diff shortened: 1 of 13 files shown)


src/Rule/Nested.php
```diff
@@ -62,32 +113,71 @@ final class Nested implements
     private iterable|null $rules;
```
arogachev, line 113: Add description.

src/Rule/Nested.php
```diff
@@ -62,32 +113,71 @@ final class Nested implements
...
 
     /**
-     * @param iterable|object|string|null $rules Rules for validate value that can be described by:
+     * @param iterable|object|string|null $rules
```
arogachev, line 116: Add description.

vjik (approved):

src/Rule/Nested.php
```diff
@@ -36,9 +37,86 @@
...
+ *
+ * ```php
+ * $rule = new Nested([
+ *     'title' => [new HasLength(max: 255)],
+ *     'author.name' => [new HasLength(min: 1)],
+ *     'files.*.url' => [new Url()],
+ * ]);
+ * ```
+ *
+ * Note that the maximum nesting level is 2, a deeper one requires wrapping with another `Nested` instance or using
```
samdark, line 73: Doc is fine but the limitation itself isn't great. Is there an issue to improve it?

src/Rule/Nested.php
```diff
@@ -36,9 +37,86 @@
...
+ *     ]),
+ * ];
+ * ```
+ *
+ * For more examples please refer to the guide.
+ *
+ * It's also possible to use DTO objects with PHP attributes, see {@see ObjectDataSet} documentation and guide for
+ * details.
+ *
+ * Supports propagation of options (see {@see PropagateOptionsHelper::propagate()} for supported options and
```
samdark, line 113: This sentence doesn't look alright. Is it possible to make it better?

arogachev: Sorry, but could you tell what exactly is wrong in your opinion?

src/Rule/Nested.php
arogachev: I was going to create one. :ok_hand:

arogachev: #508

Issue #472, opened 2022-12-19
Normalize rule names


vjik:
Normalized names:

- `AtLeast`
- `Boolean`
- `Callback`
- `Compare` (changed)
- `Composite`
- `Count`
- `Each`
- `Email`
- `Equal`
- `Greater` (changed)
- `GreaterOrEqual` (changed)
- `Length` (changed)
- `In`
- `Ip`
- `TrueValue` (changed)
- `Json`
- `Less` (changed)
- `LessOrEqual` (changed)
- `Nested`
- `NotEqual`
- `Number`
- `Regex`
- `Required`
- `StopOnError`
- `Subset`
- `Url`

ToDo:

- [ ] `CompareTo` → `Compare` (abstract class `Compare` need rename also. For example, to `AbstractCompare`)
- [ ] `GreaterThan` → `Greater`
- [ ] `GreaterThanOrEqual` → `GreaterOrEqual`
- [ ] `HasLength` → `Length`
- [ ] `IsTrue` → `TrueValue`
- [ ] `LessThan` → `Less`
- [ ] `LessThanOrEqual` → `LessOrEqual`

arogachev:
`Composite` and `StopOnError` are fine as is in my opinion. Or need to think more. We discussed why new names are not better in group chat.

arogachev:
`Boolean` is causing problems too, renamed it along with `IsTrue` in #491.

vjik:
> `Boolean` is causing problems too

What are the problems?

arogachev:
@vjik At least these:

![Screenshot from 2022-12-26 14-31-40](https://user-images.githubusercontent.com/8326201/209525495-7ed96b83-a204-4323-98de-63232fd4f2bb.png)

1. PHPDoc param, warning: "Argument type does not match the declared"
2. Argument type hint, incorrect highlighting.
3. StyleCI automatically replaces `Boolean` to `bool` causing bunch of Psalm errors.

And overall using reserved names is risky. What's safe in the current language version, can be unsafe in the updates. Suggest to rename in advance.

samdark:
My take on renamings:

- CompareTo → Compare ✅ 
- GreaterThan → Greater ❌ 
- GreaterThanOrEqual → GreaterOrEqual ❌ 
- HasLength → Length ✅
- LessThan → Less ❌ 
- LessThanOrEqual → LessOrEqual ❌

PR #519, opened 2023-01-31
Rename remaining rules, fix links in guide for built-in rules

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #472

### DIFF
--- a/src/Rule/AbstractCompare.php
+++ b/src/Rule/AbstractCompare.php
@@ -25,7 +25,7 @@ use Yiisoft\Validator\WhenInterface;
  *
  * The default comparison is based on string values, which means the values
  * are compared byte by byte. When comparing numbers, make sure to change {@see AbstractCompare::$type} to
- * {@see AbstractCompare::TYPE_NUMBER} to enable numeric comparison.
+ * {@see CompareType::NUMBER} to enable numeric comparison.
  *
  * @see CompareHandler
  *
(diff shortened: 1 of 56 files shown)


src/Rule/AbstractCompare.php
```diff
@@ -0,0 +1,277 @@
...
+ *
+ * The default comparison is based on string values, which means the values
+ * are compared byte by byte. When comparing numbers, make sure to change {@see AbstractCompare::$type} to
+ * {@see AbstractCompare::TYPE_NUMBER} to enable numeric comparison.
+ *
+ * @see CompareHandler
+ *
+ * @psalm-import-type WhenType from WhenInterface
+ */
+abstract class AbstractCompare implements
```
samdark, line 34: Makes using constants a bit uglier:

```php
Compare::TYPE_STRING;
AbstractCompare::TYPE_STRING;
```

tests/Rule/HasLengthTest.php
```diff
@@ -6,8 +6,8 @@
 
 use stdClass;
```
samdark, line 7: HasLengthTest → LengthTest?

src/Rule/AbstractCompare.php
samdark: But that's probably OK.

samdark (approved):

src/Rule/AbstractCompare.php
vjik: We can create simple class:

```php
final class CompareType {
    const STRING = 'string';
    const NUMBER = 'number';
}
```

This is resolve warning in IDE and make clearly appearance: 

```php
CompareType::STRING
```

PR #521, opened 2023-01-31
Fix float numbers equality and type casting in `Compare` rule, add datetime support, various improvements

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #520, #522, #537

Overview of additional changes:

- Fixed `Compare::$incorrectDataSetTypeMessage` to be shown on `null` values.
- Fixed `targetValueOrAttribute` parameter for `Compare::$message` and `Compare::getOptions()`. Target value was always displayed.
- Added `targetAttributeValue` error parameter for `Compare:$message` for target value extracted from attribute. Might be useful in a custom message.
- `Compare::getOptions()` - handled case for complex target value.
- Simplified test for `Compare::getOptions()`.
- Changed default for `Compare::$type` (including shortcut rules) to `CompareType::NUMBER` because it's used more.
- Added "strictly equal" / "strictly not equal" messages.

TODO:

- [x] Adjust PHPDoc.
- [x] Check / add note about very large ints that are stringified into scientific notation (https://github.com/yiisoft/validator/pull/521#issuecomment-1416703721).
- [x] Check / add note about ints that may lose precision when casted to floats (https://github.com/yiisoft/validator/pull/521#issuecomment-1416703721).


arogachev:
If the logic changes are OK, I'll add PHPDoc and more tests later.

src/Rule/CompareHandler.php
```diff
@@ -77,23 +80,68 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
...
+            '!=' => $this->assertNotEquals($type, $value, $targetValue),
+            '!==' => $this->assertNotEquals($type, $value, $targetValue, strict: true),
             '>' => $value > $targetValue,
             '>=' => $value >= $targetValue,
             '<' => $value < $targetValue,
             '<=' => $value <= $targetValue,
         };
     }
+
+    private function assertEquals(
```
samdark, line 99:
```suggestion
    private function equals(
```

because it's not a usual assertion throwing an exception on negative result of a check.

src/Rule/CompareHandler.php
```diff
@@ -77,23 +80,68 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
...
+        }
+
+        if ($type === CompareType::NUMBER) {
+            return $this->assertFloatsEqual((float) $value, (float) $targetValue);
+        }
+
+        return (string) $value == (string) $targetValue;
+    }
+
+    private function assertNotEquals(
```
samdark, line 116:
```suggestion
    private function notEquals(
```

src/Rule/CompareHandler.php
```diff
@@ -77,23 +80,68 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
...
+    ): bool {
+        if ($strict && gettype($value) !== gettype($targetValue)) {
+            return false;
+        }
+
+        if ($type === CompareType::NUMBER) {
+            return $this->assertFloatsEqual((float) $value, (float) $targetValue);
+        }
+
+        return (string) $value == (string) $targetValue;
```
samdark, line 113: Are you sure all types should be compared like that except numbers?

src/Rule/CompareHandler.php
```diff
@@ -77,23 +80,68 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
...
+    ): bool {
+        if ($strict && gettype($value) !== gettype($targetValue)) {
+            return true;
+        }
+
+        if ($type === CompareType::NUMBER) {
+            return !$this->assertFloatsEqual((float) $value, (float) $targetValue);
+        }
+
+        return (string) $value != (string) $targetValue;
```
samdark, line 130:
```suggestion
        return (string) $value !== (string) $targetValue;
```

Because types already match.

src/Rule/CompareHandler.php
```diff
@@ -77,23 +80,68 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
...
+    }
+
+    private function assertFloatsEqual(float $value, float $targetValue): bool
```
samdark, line 133:
```suggestion
    private function floatsEqual(float $value, float $targetValue): bool
```

arogachev: I renamed it, but slightly different, because suggested version doesn't imply a question.

src/Rule/CompareHandler.php
arogachev: I renamed it, but slightly different, because suggested version doesn't imply a question.

src/Rule/CompareHandler.php
arogachev: I renamed it, but slightly different, because suggested version doesn't imply a question.

src/Rule/CompareHandler.php
cebe: what is the reason to cast to string?

arogachev: That how it was in previous version. Do you have edge case examples?

src/Rule/CompareHandler.php
cebe: why is assertNotEquals() a separate implementation and not simply `return !$this->assertEquals(...)`?

src/Rule/CompareHandler.php
vjik: > That how it was in previous version. Do you have edge case examples?

For example, compare objects with `===`

vjik (commented):
Seems, need add compare type that will not do type casting.

src/Rule/CompareHandler.php
arogachev: @vjik Objects are not supported for both rule value and input value:

https://github.com/yiisoft/validator/blob/7be3488331d9348cb3e7e247f73ddac8214d5f14/src/Rule/AbstractCompare.php#L111-L112

https://github.com/yiisoft/validator/blob/7be3488331d9348cb3e7e247f73ddac8214d5f14/src/Rule/AbstractCompare.php#L111-L112

arogachev:
> Seems, need add compare type that will not do type casting.

@vjik OK, I'll add this.

src/Rule/CompareHandler.php
arogachev: @cebe Thanks. Indeed, using 1 method with negation is enough. I fixed it as well.

samdark:
Cases that could go wrong with current code:

- booleans
- DateTime
- Very large ints that are stringified into scientific notation
- ints that may lose precision when casted to floats

samdark (approved):

src/Rule/AbstractCompare.php
```diff
@@ -124,7 +124,7 @@ abstract class AbstractCompare implements
      * @psalm-param WhenType $when
      */
     public function __construct(
-        private int|float|string|bool|null $targetValue = null,
+        private mixed $targetValue = null,
         private string|null $targetAttribute = null,
         private string $incorrectInputMessage = self::DEFAULT_INCORRECT_INPUT_MESSAGE,
         private string $incorrectDataSetTypeMessage = self::DEFAULT_INCORRECT_DATA_SET_TYPE_MESSAGE,
```
arogachev, line 130: Change `type` default to `number`.

src/Rule/AbstractCompare.php
```diff
@@ -180,7 +180,7 @@ public function getTargetAttribute(): string|null
      *
      * @return string The type of the values being compared. Either {@see CompareType::STRING}
      * or {@see CompareType::NUMBER}.
-     * @psalm-return CompareType::STRING | CompareType::NUMBER $type
+     * @psalm-return CompareType::ORIGINAL | CompareType::STRING | CompareType::NUMBER $type
      *
      * @see $type
      */
```
arogachev, line 186: Separate "strictly equal" / "strictly not equal" messages.

tests/Rule/CompareTest.php
```diff
@@ -577,22 +511,248 @@ public function hasAttribute(string $attribute): bool
...
                 ],
                 [
                     'number' => [
-                        'Attribute - number, target value - , target attribute - attribute, target value or ' .
-                        'attribute - 100, value - 101.',
+                        'Attribute - number, target value - , target attribute - attribute, target attribute value ' .
+                        '- 100, target value or attribute - attribute, value - 101.',
                     ],
                 ],
             ],
```
arogachev, line 526: Check formatting of complex types.

samdark (approved):

src/Rule/CompareHandler.php
```diff
@@ -217,6 +220,8 @@ private function normalizeNumber(mixed $number): float
     {
         if ($number instanceof Stringable) {
             $number = (string) $number;
+        } elseif ($number instanceof DateTimeInterface) {
```
samdark, line 223: DateTimeInterface should be compared to DateTimeInterface only. We should not even try to compare 42 with the date.

src/Rule/CompareHandler.php
```diff
@@ -30,70 +35,196 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
...
+            $value instanceof DateTimeInterface;
+    }
+
+    /**
+     * Gets representation of the value for using with error parameter.
+     *
+     * @param string $type The type of the values being compared ({@see AbstractCompare::$type}).
+     * @psalm-param CompareType::ORIGINAL | CompareType::STRING | CompareType::NUMBER $type
+     *
+     * @param mixed $value The мalidated value.
```
vjik, line 110:
```suggestion
     * @param mixed $value The validated value.
```

src/Rule/CompareHandler.php
```diff
@@ -30,70 +35,196 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
...
-        if ($type === CompareType::NUMBER) {
-            $value = (float) $value;
-            $targetValue = (float) $targetValue;
-        } else {
-            $value = (string) $value;
-            $targetValue = (string) $targetValue;
+        if (!in_array($operator, ['==', '===', '!=', '!=='])) {
+            if ($type === CompareType::STRING) {
+                $value = (string) $value;
+                $targetValue = (string) $targetValue;
```
vjik, line 144:
```php
'target value: Unix Timestamp string, value: DateTime object with the same value, type: string, operator: ==' => [
    $dateTime->format('U'),
    [new Compare($dateTime, type: CompareType::STRING)],
],
```

Result: `Error: Object of class DateTime could not be converted to string`.

Into `checkValuesAreEqual()` also.

docs/guide/en/built-in-rules-compare.md
```diff
@@ -0,0 +1,40 @@
+# `Compare` - comparing validated value with target value
+
+## Using with `DateTime` objects
```
vjik, line 3: Is it necessary if it now works without setting the type?

vjik (changes requested):

docs/guide/en/built-in-rules-compare.md
arogachev: I think it still can be useful.

samdark (approved):

vjik (approved):

PR #542, opened 2023-02-03
Cleanup shortcut rules

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️/❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Rule/AbstractCompare.php
+++ b/src/Rule/AbstractCompare.php
@@ -133,6 +133,14 @@ abstract class AbstractCompare implements
             throw new InvalidArgumentException('Either "targetValue" or "targetAttribute" must be specified.');
         }
 
+        if ($this->type !== CompareType::NUMBER && $this->type !== CompareType::STRING) {
+            $numberType = CompareType::NUMBER;
+            $stringType = CompareType::STRING;
+            $message = "Type \"$this->type\" is not supported. The valid types are: \"$numberType\", \"$stringType\".";
+
+            throw new InvalidArgumentException($message);
+        }
+
         if (!isset($this->validOperatorsMap[$this->operator])) {
             $wrapInQuotesCallable = static fn (string $operator): string => '"' . $operator . '"';
             /** @var string[] $validOperators */
--- a/src/Rule/AbstractCompare.php
+++ b/src/Rule/AbstractCompare.php
@@ -14,6 +14,8 @@ use Yiisoft\Validator\SkipOnEmptyInterface;
 use Yiisoft\Validator\SkipOnErrorInterface;
 use Yiisoft\Validator\WhenInterface;
 
+use function in_array;
+
 /**
  * Abstract base for all the comparison validation rules.
  *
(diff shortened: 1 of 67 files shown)


tests/ValidatorTest.php
```diff
@@ -283,9 +283,9 @@ public function testDiverseTypes($dataSet): void
     public function testNullAsDataSet(): void
     {
         $validator = new Validator();
-        $result = $validator->validate(null, ['property' => [new Compare(null)]]);
```
vjik, line 286: We can keep this test and add new.

src/Rule/AbstractCompare.php
```diff
@@ -110,16 +120,27 @@ abstract class AbstractCompare implements
...
         private string $operator = '==',
         private mixed $skipOnEmpty = null,
         private bool $skipOnError = false,
         private Closure|null $when = null,
     ) {
+        if ($this->targetValue === null && $this->targetAttribute === null) {
+            throw new InvalidArgumentException('Either "targetValue" or "targetAttribute" must be specified.');
+        }
+
+        if ($this->type !== CompareType::NUMBER && $this->type !== CompareType::STRING) {
```
vjik, line 136: Need fix psalm error

vjik (changes requested):

tests/ValidatorTest.php
arogachev: It no longer can be kept after these changes:

https://github.com/yiisoft/validator/blob/7dfab7f12b01300841c7d5fae035a5d1682672c2/src/Rule/AbstractCompare.php#L132-L134

Would be the same as:

https://github.com/yiisoft/validator/blob/67313f798abf79801e123aa1262d88ff5b5d09be/tests/Rule/CompareTest.php#L23-L29

arogachev: A more in-depth explanation and update.

Actually, initially I found that this check was present in shortcuts but missing in base rule:

https://github.com/yiisoft/validator/blob/7dfab7f12b01300841c7d5fae035a5d1682672c2/src/Rule/AbstractCompare.php#L132-L134

This implies 2 problems:

- Copy paste.
- Inconsistency in logic because both `Compare` and shortcuts are available for usage.

I moved it to the base rule, but this revealed another problem - not being able to specify `null` as `targetValue` (previously it was the case too, but only for shortcuts). Then i figured it out that this check is not needed at all. Precedence of `targetValue` is handled and documented.

I reverted according changes, so, at the end, this test can be kept.

src/Rule/AbstractCompare.php
```diff
@@ -259,4 +277,21 @@ public function getHandler(): string
...
+     *
+     * @param string[] $items Initial list of strings to format.
+     *
+     * @return string Resulting formatted string.
+     */
+    private function getQuotedList(array $items): string
+    {
+        $wrapInQuotesCallable = static fn (string $item): string => '"' . $item . '"';
+
+        return implode(', ', array_map($wrapInQuotesCallable, $items));
```
samdark, line 295:
```suggestion
        return '"'. implode('", "', array_map($wrapInQuotesCallable, $items)) . '"';
```

samdark (approved):

src/Rule/AbstractCompare.php
arogachev: Should be:

```php
return '"'. implode('", "', $items) . '"';
```

But I got the idea and applied it.

src/Rule/AbstractCompare.php
```diff
@@ -41,11 +43,28 @@ abstract class AbstractCompare implements
...
+     */
+    protected const DEFAULT_INCORRECT_DATA_SET_TYPE_MESSAGE = 'The attribute value returned from a custom data set ' .
+    'must have a scalar type.';
+
+    /**
+     * @var string[] List of valid types.
+     *
+     * @see CompareType
+     */
+    private static array $validTypes = [CompareType::STRING, CompareType::NUMBER];
```
vjik, line 61:
```suggestion
    private const VALID_TYPES = [CompareType::STRING, CompareType::NUMBER];
```

src/Rule/AbstractCompare.php
```diff
@@ -41,11 +43,28 @@ abstract class AbstractCompare implements
...
     /**
      * @var array Map of valid operators.
      * It's used instead of a list for better performance.
+     * @psalm-var array<string, 1>
      */
-    private array $validOperatorsMap = [
+    private static array $validOperatorsMap = [
```
vjik, line 67:
```suggestion
    private const VALID_OPERATORS_MAP = [
```

src/Rule/AbstractCompare.php
samdark: Why was it marked as resolved?

samdark (approved):
Overall alright but need to address @vjik comment properly.

samdark (approved):

src/Rule/AbstractCompare.php
arogachev: The first time the push was unsuccessful. Now it's OK.

vjik (approved):

PR #550, opened 2023-02-11
Use `RulesProviderInterface` in complex rules (`Nested`, `Each`, `Composite`, `StopOnError`)

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/src/Rule/Composite.php
+++ b/src/Rule/Composite.php
@@ -106,7 +106,7 @@ class Composite implements
     protected Closure|null $when = null;
 
     /**
-     * @param iterable|callable|RuleInterface $rules A set of rules that needs to be grouped. They will be normalized
+     * @param callable|iterable|RuleInterface $rules A set of rules that needs to be grouped. They will be normalized
      * using {@see RulesNormalizer}.
      * @psalm-param iterable<Closure|RuleInterface> $rules
      *
--- a/src/Rule/Composite.php
+++ b/src/Rule/Composite.php
@@ -86,7 +86,7 @@ class Composite implements
 
     /**
      * @var iterable A set of normalized rules that needs to be grouped.
-     * @psalm-var iterable<int, RuleInterface>
+     * @psalm-var iterable<iterable<RuleInterface>|RuleInterface>
      */
     protected iterable $rules = [];
     /**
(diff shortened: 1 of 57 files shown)


vjik (commented):
Will be good add more specific psalm annotation for `$rules` parameter in `RulesProviderInterface::getRules()` method.

samdark (approved):

src/Validator.php
```diff
@@ -24,7 +24,7 @@
  * The only built-in implementation of {@see ValidatorInterface}, the main class / entry point processing all the data
  * and rules with validation context together and performing the actual validation.
  *
- * @psalm-import-type RulesType from ValidatorInterface
+ * @psalm-import-type RawRulesType from ValidatorInterface
```
vjik, line 27: This type don't used in `Validator` class

src/Rule/Composite.php
```diff
@@ -117,8 +118,8 @@ class Composite implements
      * @psalm-param WhenType $when
      */
     public function __construct(
-        iterable $rules = [],
-        bool|callable|null $skipOnEmpty = null,
+        iterable|callable|RuleInterface $rules = [],
```
vjik, line 121: Why create composite rule with one rule (callable or RuleInterface) ?

src/Rule/Nested.php
```diff
@@ -220,7 +219,7 @@ final class Nested implements
      * @psalm-param WhenType $when
      */
     public function __construct(
-        iterable|object|string|null $rules = null,
+        iterable|object|string $rules = [],
```
vjik, line 222: `null` need for use attributes from attributes of validated object

vjik (changes requested):

src/Rule/Nested.php
arogachev: Because `RulesProviderInterface::getRules()` return type doesn't allow `null` values:

https://github.com/yiisoft/validator/blob/db5d9e3564ec30e5b8c0cc9d7963a721f79c31d1/src/RulesProviderInterface.php#L15

The handler is modified accordingly.

vjik: Means `Nested` must not implement `RulesProviderInterface`.

vjik (approved):

PR #551, opened 2023-02-11
Compare `DateTimeInterface` by default

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  |

### DIFF
--- a/tests/Rule/CompareTest.php
+++ b/tests/Rule/CompareTest.php
@@ -250,6 +250,10 @@ final class CompareTest extends RuleTestCase
                 $dateTime,
                 [new Compare($dateTime, type: CompareType::ORIGINAL)],
             ],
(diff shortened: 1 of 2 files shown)


samdark (approved):

tests/Rule/CompareTest.php
```diff
@@ -250,6 +250,10 @@ public function __toString(): string
                 $dateTime,
                 [new Compare($dateTime, type: CompareType::ORIGINAL)],
             ],
+            'target value: DateTime object, value: the same DateTime object, type: default, operator: ===' => [
```
arogachev, line 253:
```suggestion
            'target value: Unit TimeStamp string, value: the DateTime object with the same value, type: default, operator: ===' => [
```

arogachev (approved):

Issue #558, opened 2023-02-13
Getting access to attributes out of nested validation context


bouledepate:
Hello. I have a problem when I try to write conditional validation using ValidationContext object and trying to get attribute value, which not included into Nested collection. For example, I have this:
```(php)
$rules = [
    'method' => [new Required()],
    'attributes' => new Nested([
        'abc' => [new Required(when: static function (mixed $value, ValidationContext $context): bool {
            $method = $context->getDataSet()->getAttributeValue('method');
            return $method === Method::POST->value;
        }),
        ...
    ])
];
```
When I try to get value via $context->getDataSet()->getAttributeValue('method'), I got null. But If this condition will not be in Nested collection, I receive needed value.

Documentation says: "$context->getDataSet()->getAttributeValue() method allows you to get any other attribute's value within the ValidationContext". Does it mean I must have access those attributes from Nested ValidationContext or not? Or what kind of attributes must be included into validation context?

arogachev:
Thanks for reporting! Fixed in #559.

PR #559, opened 2023-02-13
Add isolated data set to validation context

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #558

[Diff](https://github.com/yiisoft/validator/compare/master...context-improve)


src/ValidationContext.php
```diff
@@ -125,12 +129,12 @@ public function validate(mixed $data, callable|iterable|object|string|null $rule
     {
         $this->requireValidator();
 
-        $currentDataSet = $this->dataSet;
+        $currentIsolatedDataSet = $this->isolatedDataSet;
```
arogachev, line 132: Need tests.

src/ValidationContext.php
```diff
@@ -39,10 +39,12 @@ final class ValidationContext
 
     /**
      * @var DataSetInterface|null Data set the attribute belongs to.
-     * `null` if data set was not set with {@see setDataSet()} yet.
+     * `null` if data set was not set with {@see setIsolatedDataSet()} yet.
```
arogachev, line 42: The description is not accurate. `$dataSet` is not set with this method.

src/ValidationContext.php
```diff
@@ -39,10 +39,12 @@ final class ValidationContext
...
      */
     private ?DataSetInterface $dataSet = null;
 
+    private ?DataSetInterface $isolatedDataSet = null;
```
arogachev, line 46: Need to add PHPDoc.

src/ValidationContext.php
```diff
@@ -80,7 +82,8 @@ public function __construct(
     public function setContextDataOnce(
         ValidatorInterface $validator,
         AttributeTranslatorInterface $attributeTranslator,
-        mixed $rawData
+        mixed $rawData,
+        DataSetInterface $dataSet
```
arogachev, line 86: Need to update PHPDoc.

src/ValidationContext.php
arogachev: And mention the difference between `$dataSet` and `$isolatedDataSet` (in the other places as well).

src/ValidationContext.php
```diff
@@ -149,18 +153,24 @@ public function getRawData(): mixed
         return $this->rawData;
     }
 
+    public function getDataSet(): DataSetInterface
```
arogachev, line 156: PHPDoc is missing.

arogachev (changes requested):

src/Rule/CompareHandler.php
```diff
@@ -42,7 +42,7 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
 
         if ($targetValue === null && $targetAttribute !== null) {
             /** @var mixed $targetValue */
-            $targetValue = $context->getDataSet()->getAttributeValue($targetAttribute);
+            $targetValue = $context->getIsolatedDataSet()->getAttributeValue($targetAttribute);
```
arogachev, line 45: Without this change the tests will still pass.

arogachev:
I can help with PHPDoc.

src/Rule/CompareHandler.php
arogachev: Still actual.

arogachev: Will be handled separately - #563.

arogachev (approved):

Issue #508, opened 2023-01-11
`Nested` rule - allow deeper level of nesting without using other instances


arogachev:
https://github.com/yiisoft/validator/blob/27e6c3c100340d1adbf622d352bbbd469e44d9b4/src/Rule/Nested.php#L73-L94

Issue #479, opened 2022-12-20
Support nested arrays in `PropagateOptionsHelper::propagate()`


arogachev:
For example:

```php
new Nested([
    [
        'data' => [
            'name' => [new HasLength(min: 5)],
            'age' => [
                new Number(min: 18),
                new Number(min: 20),
            ],
        ],
    ],
]);
```

PR #564, opened 2023-02-14
Allow any deep level in `Nested` rule

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #508, #479

### DIFF
--- a/docs/guide/en/built-in-rules-nested.md
+++ b/docs/guide/en/built-in-rules-nested.md
@@ -81,7 +81,7 @@ Some code can be omitted for brevity.
 
 #### Inner `Nested` instances
 
-Inner `Nested` instances can be omitted, but only for nesting levels no greater than 2:
+Inner `Nested` instances can be omitted:
 
 ```php
 use Yiisoft\Validator\Rule\Length;
--- a/docs/guide/en/built-in-rules-nested.md
+++ b/docs/guide/en/built-in-rules-nested.md
@@ -94,24 +94,6 @@ $rule = new Nested([
 ]);
 ```
 
-This will not work:
-
-```php
-use Yiisoft\Validator\Rule\Length;
-use Yiisoft\Validator\Rule\Nested;
-
-$rule = new Nested([
-    'author' => [
-        'name' => [
-            'surname' => [new Length(min: 1)],
-        ],
-    ],
-]);
-```
-
-This limitation is planned to be removed in the future. In order for this example to work, it needs to be rewritten
-to be wrapped with another `Nested` instance or using the short syntax shown above.
-
 #### Inner arrays for single rules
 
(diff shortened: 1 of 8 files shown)


docs/guide/en/built-in-rules-nested.md
```diff
@@ -81,7 +81,7 @@ Some code can be omitted for brevity.
 
 #### Inner `Nested` instances
 
-Inner `Nested` instances can be omitted, but only for nesting levels no greater than 2:
+Inner `Nested` instances can be omitted:
```
arogachev, line 84: The section below needs to be removed (starting from "This will not work:").

src/Rule/Nested.php
```diff
@@ -364,25 +325,59 @@ private function prepareRules(iterable|object|string|null $source): void
...
     }
 
+    /**
+     * @psalm-param array<array<RuleInterface>|RuleInterface> $rawRules
+     * @psalm-param array<list<RuleInterface>|RuleInterface> $result
+     */
+    private function prepareRules(array $rawRules, array &$result, ?string $baseValuePath = null): void
+    {
+        foreach ($rawRules as $valuePath => $validationRules) {
+            if (is_int($valuePath)) {
```
arogachev, line 350: [EA] Using '\is_int(...)' would enable some of opcode optimizations.

src/Rule/Nested.php
```diff
@@ -364,25 +325,59 @@ private function prepareRules(iterable|object|string|null $source): void
...
+                $key = $baseValuePath;
+            } else {
+                $key = ($baseValuePath !== null ? $baseValuePath . '.' : '') . $valuePath;
+            }
+
+            if (is_array($validationRules)) {
```
arogachev, line 356: [EA] Using '\is_array(...)' would enable some of opcode optimizations.

src/Rule/Nested.php
```diff
@@ -342,11 +304,10 @@ public function getNoPropertyPathMessage(): string
      * @throws InvalidArgumentException When rules' source has wrong type.
      * @throws InvalidArgumentException When source contains items that are not rules.
      */
-    private function prepareRules(iterable|object|string|null $source): void
+    private function setRules(iterable|object|string|null $source): void
```
arogachev, line 307: Maybe rename to something like `initRules`? For me, `set` prefix is more associated with public setters.

arogachev: Have a better idea for naming in #568.

arogachev (changes requested):
The changes in #568 need to be addressed.

arogachev (approved):

PR #566, opened 2023-02-14
Use validator rules format in Each

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  |

PR #569, opened 2023-02-15
Corrections for PR #566

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | 

Corrections for PR #566.

### DIFF
--- a/src/Rule/Each.php
+++ b/src/Rule/Each.php
@@ -58,7 +58,7 @@ use Yiisoft\Validator\WhenInterface;
  * @see EachHandler Corresponding handler performing the actual validation.
  *
  * @psalm-import-type RulesTypeWithoutNull from ValidatorInterface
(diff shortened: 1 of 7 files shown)


arogachev:
I tried to add null support, but it turned out that it is harder than I initially thought. Suggest to handle it later. Maybe it's not needed at all.

src/Rule/Each.php
```diff
@@ -125,7 +128,7 @@ public function getName(): string
     public function propagateOptions(): void
     {
         foreach ($this->rules as $key => $rules) {
-            $this->rules[$key] = PropagateOptionsHelper::propagate($this, $rules);
+            $this->rules[$key] = new RulesNormalizerIterator(PropagateOptionsHelper::propagate($this, $rules));
```
vjik, line 131: Why create iterator from array?

PR #573, opened 2023-02-21
Fix mutant in `Nested::propagateOptions()`, add and simplify tests

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |


vjik (changes requested):
https://github.com/yiisoft/validator/pull/575

PR #576, opened 2023-02-21
Fix new mutants

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  |


src/Validator.php
```diff
@@ -112,7 +112,6 @@ public function validate(
             } else {
                 /** @psalm-suppress MixedAssignment */
                 $validatedData = $dataSet->getAttributeValue($attribute);
-                $context->setParameter(ValidationContext::PARAMETER_VALUE_AS_ARRAY, null);
```
vjik, line 115: It is need.

vjik (changes requested):

src/Validator.php
arogachev: It's not necessary because of the usage of `ArrayHelper` with default value in `getParameter()`:

https://github.com/yiisoft/validator/blob/4d0c9407118f76a641908fcadd19b5557ceab381/src/ValidationContext.php#L275

arogachev: Fixed it differently.

src/Rule/EachHandler.php
```diff
@@ -23,8 +23,9 @@ public function validate(mixed $value, object $rule, ValidationContext $context)
             throw new UnexpectedRuleException(Each::class, $rule);
         }
 
+        $valueAsArray = $context->getParameter(ValidationContext::PARAMETER_VALUE_AS_ARRAY);
         /** @var mixed $value */
-        $value = $context->getParameter(ValidationContext::PARAMETER_VALUE_AS_ARRAY) ?? $value;
+        $value = $valueAsArray !== null ? $valueAsArray : $value;
```
vjik, line 28: Seems, it's same but with more code and more complicated?

arogachev: Reverted the change.

vjik (approved):

Issue #431, opened 2022-12-07
Use `WhenEmpty` with other empty countables


arogachev:
Some rules like `Count` allows iterator as input value which can be empty too (with some condition or intentionally maybe). Currently only direct check for an empty array is used:

https://github.com/yiisoft/validator/blob/09fed8119fa3ce695e81a2776448e28bc108421e/src/EmptyCriteria/WhenEmpty.php#L21

The tests for this case won't be redundant anyway.

PR #586, opened 2023-02-24
Skip empty iterable

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌
| Fixed issues  | #431

Additional changes:

- Fix usages of `new Count(exactly: ...)` (an addition to #498).
- Allow `0` as a limit in `CountableLimitTrait`.

### DIFF
--- a/src/Rule/Trait/CountableLimitTrait.php
+++ b/src/Rule/Trait/CountableLimitTrait.php
@@ -123,11 +123,11 @@ trait CountableLimitTrait
         }
 
         if (
-            ($this->min !== null && $this->min <= 0) ||
-            ($this->max !== null && $this->max <= 0) ||
-            ($this->exactly !== null && $this->exactly <= 0)
+            ($this->min !== null && $this->min < 0) ||
+            ($this->max !== null && $this->max < 0) ||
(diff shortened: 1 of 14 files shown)


src/Rule/Trait/CountableLimitTrait.php
```diff
@@ -123,21 +123,25 @@ private function initCountableLimitProperties(
...
             if ($this->min > $this->max) {
                 throw new InvalidArgumentException('$min must be lower than $max.');
             }
 
             if ($this->min === $this->max) {
-                throw new InvalidArgumentException('Use $exactly instead.');
+                throw new InvalidArgumentException($messageForUsingExactly);
             }
+        } elseif ($this->min === null && $this->max === 0) {
+            throw new InvalidArgumentException($messageForUsingExactly);
```
vjik, line 144: Max count equal to `0` is OK.  It's value can by dynamic:

`new Count(max: $max)` better than `new Count(max: $max> 0 ? $max : null, exactly: $max === 0 ? 0 : null)`.

vjik (changes requested):

samdark (approved):

Issue #598, opened 2023-04-03
Ability add base result before validation


vjik:
My suggestion: add ability add base result to `Context` and use [here](https://github.com/yiisoft/validator/blob/b0044e016768cdfa40d451612da054d6d3bf2162/src/Validator.php#L122).

Use case: input model with prevalidation raw data.

samdark:
What is prevalidation data?

vjik:
Raw values and populated values can be different. My service validate raw data into populate process and then run classic validation. 

For example:

```php
final class SimpleModel implements ValidatedModelInterface
{
    use ValidatedModelTrait;

    public function __construct(
        #[PreValidate(new Required())]
        #[Length(min: 3, skipOnError: true)]
        public string $a = '.',
        public string $b = '.',
        public string $c = '.',
    ) {
    }
}

$service->create(SimpleModel::class, data: []);
```

In raw data `a` not exists and `Required` add error. Then length should not add error because option `skipOnError` is true.

BoShurik:
Why don't you use separate models: for user input (with raw data) and model with populated data (converted from user input)?

vjik:
> Why don't you use separate models: for user input (with raw data) and model with populated data (converted from user input)?

Not convient. One input model is simple.

samdark:
It is a lazy approach that is convenient at first but:

1. It complicates validator.
2. The use-case is not common enough.
3. With application growth these two layer inputs are likely to diverge more and such approach will create more problems than it is trying to solve.

Overall, I don't think it is a good idea.

vjik:
Decided don't make it, because use-case is changed: if pre validation is fail then not run validation.

PR #599, opened 2023-04-04
Add ability to pass predefined result

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #598

### DIFF
--- a/docs/guide/en/result.md
+++ b/docs/guide/en/result.md
@@ -281,3 +281,23 @@ $result->getAttributeErrors('email');
 ```
 
 [Using keys containing separator / shortcut]: #using-keys-containing-separator--shortcut
+
+## Predefined result
+
+Sometimes you can have information about validation errors before validation process. In this case use predefined result
+and pass it to validator via parameter `ValidationContext::PARAMETER_PREDEFINED_RESULT` of validation context:
+
+```php
+use Yiisoft\Validator\Result;
+use Yiisoft\Validator\Validator;
+use Yiisoft\Validator\ValidationContext;
+
+$predefinedResult = new Result();
+$predefinedResult->addError('Invalid value.', valuePath: ['key']);
+
+$context = new ValidationContext([
+    ValidationContext::PARAMETER_PREDEFINED_RESULT => $predefinedResult,
+])
+
+$result = (new Validator)->validate($model, context: $context);
+```
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.1.0 under development
 
+- New #598: Add ability to pass predefined result (@vjik)
 - Enh #594: Add `StringValue` rule (@arogachev)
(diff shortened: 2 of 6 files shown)


docs/guide/en/result.md
```diff
@@ -281,3 +281,23 @@ $result->getAttributeErrors('email');
...
+use Yiisoft\Validator\ValidationContext;
+
+$predefinedResult = new Result();
+$predefinedResult->addError('Invalid value.', valuePath: ['key']);
+
+$context = new ValidationContext([
+    ValidationContext::PARAMETER_PREDEFINED_RESULT => $predefinedResult,
+])
+
+$result = (new Validator)->validate($model, context: $context);
```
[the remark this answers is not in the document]
vjik: I considered this option. It's not work in this case:

```php
final class SimpleModel implements ValidatedModelInterface
{
    use ValidatedModelTrait;

    public function __construct(
        #[PreValidate(new Required())]
        #[Length(min: 3, skipOnError: true)]
        public string $a = '.',
        public string $b = '.',
        public string $c = '.',
    ) {
    }
}

$service->create(SimpleModel::class, data: []);
```

`skipOnError` options work incorrect, because validator not know about errors on prevalidation level.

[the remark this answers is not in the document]
samdark: OK. For `skipOnError`, indeed, order matters.

docs/guide/en/result.md
```diff
@@ -281,3 +281,23 @@ $result->getAttributeErrors('email');
 ```
 
 [Using keys containing separator / shortcut]: #using-keys-containing-separator--shortcut
+
+## Predefined result
+
+Sometimes you can have information about validation errors before validation process. In this case use predefined result
+and pass it to validator via parameter `ValidationContext::PARAMETER_PREDEFINED_RESULT` of validation context:
+
```
samdark, line 289:
```suggestion
Sometimes you can have multiple stages of validation. For example, validating the raw data, normalizing it, validating normalized data. In these cases you need validation errors from previous step validation process. You can use predefined result
and pass it to validator via parameter `ValidationContext::PARAMETER_PREDEFINED_RESULT` of validation context:

```

samdark: This way intent is cleaner.

docs/guide/en/result.md
```diff
@@ -281,3 +281,23 @@ $result->getAttributeErrors('email');
 ```
 
 [Using keys containing separator / shortcut]: #using-keys-containing-separator--shortcut
+
+## Predefined result
+
+Sometimes you can have information about validation errors before validation process. In this case use predefined result
+and pass it to validator via parameter `ValidationContext::PARAMETER_PREDEFINED_RESULT` of validation context:
```
samdark, line 288: How about naming it ValidationContext::PREVIOUS_RESULT?

vjik:
See https://github.com/yiisoft/validator/issues/598#issuecomment-1497276408

Issue #610, opened 2023-06-22
Validation error output


bouledepate:
Hi everyone again :)

My input data array contains dot-separated keys. According to the documentation, it is recommended to escape such keys so as not to break them into subcollections. This is good, but when I get the list of errors via `getErrorMessagesIndexedByPath` I don't want to see the escaping. I looked at the code and saw what can be done:

The 1st way is to move the parameter to a separate flag.
```php
    public function getErrorMessagesIndexedByPath(string $separator = '.', bool $escape = true): array
    {
        $errors = [];
        foreach ($this->errors as $error) {
            $stringValuePath = implode($separator, $error->getValuePath($escape));
            $errors[$stringValuePath][] = $error->getMessage();
        }
        return $errors;
    }
```
2nd way - just remove it.
```php
    public function getErrorMessagesIndexedByPath(string $separator = '.'): array
    {
        $errors = [];
        foreach ($this->errors as $error) {
            $stringValuePath = implode($separator, $error->getValuePath());
            $errors[$stringValuePath][] = $error->getMessage();
        }
        return $errors;
    }
```
Example. If the parameter is set to "true", then I get this output. 
```json
    "error": {
        "code": 790,
        "title": "Validation error!",
        "detail": {
            "filters.amount\\.value": [
                "Value must be a number."
            ]
        }
    }
```
Otherwise like this.
```json
"error": {
        "code": 790,
        "title": "Validation error!",
        "detail": {
            "filters.amount.value": [
                "Value must be a number."
            ]
        }
    }
```
Do you agree that the second conclusion is much better and more convenient? Thank you!

### What steps will reproduce the problem?
1. Make validation of data with keys 'amount.value' for example.
2. Request the error messages by 'getErrorMessagesIndexedByPath';

### What is the expected result?
Paths without escaping.

### What do you get instead?
Paths with escaping.

### Additional info

| Q                | A
| ---------------- | ---
| Version          | 1.0.1
| PHP version      |  8.2.6
| Operating system | Linux Ubuntu 20.04

arogachev:
What is your use case for this method?

bouledepate:
@arogachev, 

I have REST API. After validating the data with this method, I get all the errors given their path, since I'm using a `Nested` validator. Then I generate an exception, where I put the received array of errors, encode it through `json_encode` and give it to the client. And as planned, the key in this array can look like: "account.profile.name".
 
[...]

PR #612, opened 2023-06-22
Disable asterisk char escaping + Add `$escape` parameter to several methods of `Result`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  |  #610

### DIFF
--- a/src/Result.php
+++ b/src/Result.php
@@ -171,8 +171,7 @@ final class Result
         string $attribute,
         string $separator = '.',
         ?string $escape = '.',
-    ): array
-    {
+    ): array {
         $errors = [];
         foreach ($this->errors as $error) {
             $firstItem = $error->getValuePath()[0] ?? '';
--- a/src/Result.php
+++ b/src/Result.php
@@ -82,7 +82,7 @@ final class Result
     {
         $errors = [];
         foreach ($this->errors as $error) {
-            $stringValuePath = implode($separator, $error->getValuePath(true));
+            $stringValuePath = implode($separator, $error->getValuePath($separator));
             $errors[$stringValuePath][] = $error->getMessage();
         }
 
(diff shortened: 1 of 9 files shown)


src/Result.php
```diff
@@ -158,21 +160,26 @@ public function getAttributeErrorMessages(string $attribute): array
      *
      * @param string $attribute Attribute name.
      * @param string $separator Attribute path separator. Dot is used by default.
+     * @param string|null $escape Symbol that will be escaped with a backslash char (`\`) into path elements.
+     * When is null value path returns without escaping.
```
samdark, line 164:
```suggestion
     * @param string|null $escape Symbol that will be escaped with a backslash char (`\`) in path elements.
     * When it's null path is returned without escaping.
```

src/Result.php
```diff
@@ -73,16 +73,18 @@ public function getErrorMessages(): array
      * Each value is an array of error message strings.
      *
      * @param string $separator Attribute path separator. Dot is used by default.
+     * @param string|null $escape Symbol that will be escaped with a backslash char (`\`) into path elements.
+     * When is null value path returns without escaping.
```
samdark, line 77:
```suggestion
     * @param string|null $escape Symbol that will be escaped with a backslash char (`\`) in path elements.
     * When it's null path is returned without escaping.
```

tests/ErrorTest.php
```diff
@@ -0,0 +1,66 @@
...
+
+        $this->assertSame($expectedValuePath, $error->getValuePath($escape));
+    }
+
+    public function testTooLongEscapeSymbol(): void
+    {
+        $error = new Error('');
+
+        $this->expectException(InvalidArgumentException::class);
+        $this->expectExceptionMessage('Escape symbol must contain exactly one character.');
```
samdark, line 63:
```suggestion
        $this->expectExceptionMessage('Escape symbol must be exactly one character.');
```

samdark (approved):

Issue #630, opened 2023-11-07
Include attribute name in error messages when it's present


arogachev:
```
Name must contain at least 8 characters.
```

instead of

```
This value must contain at least 8 characters.
```

Suggested by @vjik.

dood-:
I'll take it

arogachev:
I had an idea regarding to this issue. We discussed it within the team, but I'm not sure it was conveyed / understood. The suggestion of mine is to not have separate error messages, but instead, reorganize existing ones and make them dynamic. 

For example let's take URL rule's error message:

https://github.com/yiisoft/validator/blob/e18482f3908322bcc49b4ae480dca282792e8e9e/src/Rule/Url.php#L83

It can be transformed like this:

```
{something} is not a valid URL.
```

{something} is a new placeholder (working title).

If attribute display is set (can be done with a separate setting / flag) and attribute name is "Vendor site", the message becomes:

```
Vendor site is not a valid URL.
```

Otherwise:

```
This value is not a valid URL.
```

"This value" is the most used definition of a validated value in the current error messages. Can be customizable too.

samdark:
Related to https://github.com/yiisoft/validator/pull/643

dood-:
I think it is posiiblie to add the "error_message_template" property in the config and use it:
`'error_message_template' => 'This value {message}!'` or `'error_message_template' => '{attribute} {message}'`.

It can be overridden using the attribute:
```
#[ErrorMessage(template: 'This value {message}')] // Redefine the template for the entire object
final class Dto
{
    #[Required]
    #[Label('First Name')]
    #[ErrorMessage(template: '{attribute} {message}')] // or for a specific property
    public string $name = '';

    #[Number(min: 21)]
    public int $age = 17;
}
```

vjik:
I thnik, @arogachev suggest modify current error messages. For example:

```
The value must be a string.
↓
{attribute} must be a string.
```

By default message will be "Value must be a string.".
If exist label (for example, "Name"), message will be "Name must be string".

@arogachev, is it right?

arogachev:
@vjik Yes.

PR #644, opened 2024-01-02
Include attribute name in error messages when it's present

dood-:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  | #630


docs/guide/en/configuring-rules-via-php-attributes.md
```diff
@@ -74,7 +74,7 @@ final class User
 
 > **Note:** [readonly properties] are supported only starting from PHP 8.1.
 
-Error messages may include an `{attribute}` placeholder that is replaced with the name of the property. If you would 
+Error messages may include an `{label}` placeholder that is replaced with the name of the property. If you would
```
samdark, line 77:
```suggestion
Error messages may include a `{label}` placeholder that is replaced with the name of the property. If you would 
```

docs/guide/en/configuring-rules-via-php-attributes.md
```diff
@@ -91,6 +91,22 @@ final class User
...
 ```
 
+Alternatively, you can specify a label for all class properties:
+
+```php
+use Yiisoft\Validator\Label;
+use Yiisoft\Validator\Rule\Length;
+use Yiisoft\Validator\Rule\Required;
+
+#[Label('This value')]
```
samdark, line 101: Why? What's the use-case for that?

samdark (changes requested):

docs/guide/en/configuring-rules-via-php-attributes.md
samdark: Like "**The value** should be 42"?

dood-: "The value should be 42" by default. 
If a class or property has #[Label], then the names of attributes are used

https://github.com/yiisoft/validator/blob/1e73321d83d197b0effb16e5d2c32a95d60913e4/tests/ValidatorTest.php#L1464-L1509

samdark (approved):

dood-:
Since I added a new placeholder `{label}`, perhaps it is better to make the placeholder `{attribute}` always be replaced by the name of the property?

samdark:
@vjik, @arogachev what do you think?

vjik (commented):
Hm. Is there a case when need attribute in message and at the same time, there is a label?

Seems, it's overhead. I suggest use placeholder `{attribute}` only and return translated label, if it exist, or translated attribute otherwise.

samdark:
No, I can't imagine such a case and I agree with @vjik. @dood- would you please adjust the name to be `{attribute}`? Thanks.

dood-:
@samdark, @vjik If we leave "The value" as a default value, then there is a problem with displaying the attribute names of arrays.

Here is a test for an example:
https://github.com/yiisoft/validator/blob/1e73321d83d197b0effb16e5d2c32a95d60913e4/tests/Rule/OneOfTest.php#L172-L180

We can not specify a Label attribute for the array or implement the LabelsProviderInterface, so the placeholder value will always be "The value"

vjik:
> If we leave "The value" as a default value

There will be no default value.

Will be one attribute `{attribute}`. Value of this placeholder by priority:

1) Translated label (if label exist).
2) Translated attribute.

For case with arrays always will be used translated attribute.

samdark (approved):

samdark:
> If necessary, I can remove Attribute::TARGET_CLASS from the Label attribute.

Yes, makes sense.

tests/Rule/AtLeastTest.php
```diff
@@ -199,71 +199,71 @@ public function dataValidationFailed(): array
...
-                $class,
-                [new AtLeast(['attr1', 'attr2'], min: 2, message: 'Custom message.')],
-                ['' => ['Custom message.']],
-            ],
-            'custom message with parameters' => [
-                $class,
-                [new AtLeast(['attr1', 'attr2'], min: 2, message: 'Attributes - {attributes}, min - {min}.')],
-                ['' => ['Attributes - "attr1", "attr2", min - 2.']],
-            ],
+            //'incorrect input' => [
```
vjik, line 202: Does it need to uncomment?

src/ValidationContext.php
```diff
@@ -248,11 +248,7 @@ public function getAttributeLabel(): ?string
      */
     public function getTranslatedAttribute(): ?string
     {
-        if ($this->attribute === null) {
-            return null;
-        }
-
-        $label = $this->attributeLabel ?? $this->attribute;
+        $label = $this->attributeLabel ?? $this->attribute ?? 'The value';
```
vjik, line 251: May be use `Value` without `The`? 
It's will be more consistent with attribute names, for example `age`.

Also, suggest for attributes up first symbold, for exmaple `age` → `Age`.

It would be ideal to be able to use several tokens:
• `{attribute}` — `age` or `value`
• `{Attribute}` — `Age` or `Value`

It allow to create pretty message.

vjik (changes requested):

dood-:
> Also need kill mutants from mutation testing.

Is it possible to disable mutants for a specific php version or just use @infection-ignore-all?

samdark:
I guess @vjik meant fixing tests so there are no mutants, not ignoring these.

src/Rule/DateTimeHandler.php
vjik, line 1: In `DateTime` rule need replace `The {attribute}` to `{Attribute}` into messages.

messages/ru/yii-validator.php
vjik, line 1: Need adapt other translations also.

tests/Rule/DateTimeTest.php
```diff
@@ -63,42 +63,42 @@ public function dataValidationFailed(): array
...
                 'datetime',
                 [new DateTime(message: 'Attribute - {attribute}, value - {value}.')],
-                ['' => ['Attribute - , value - datetime.']],
+                ['' => ['Attribute - value, value - datetime.']],
             ],
             'empty string and custom message' => [
                 '',
                 [new DateTime()],
-                ['' => ['The  must be a date.']],
+                ['' => ['The value must be a date.']],
```
vjik, line 91: Need to replace `The {attribute}` to `{Attribute}` in `DateTime` error messages

vjik (changes requested):
Overall looks good 👍
There's just a little bit left to finish.

vjik (approved):

arogachev:
Thanks for you great work! I will review this as well.

src/ValidationContext.php
```diff
@@ -243,16 +243,11 @@ public function getAttributeLabel(): ?string
...
      */
-    public function getTranslatedAttribute(): ?string
+    public function getTranslatedAttribute(): string
     {
-        if ($this->attribute === null) {
-            return null;
-        }
-
-        $label = $this->attributeLabel ?? $this->attribute;
+        $label = $this->attributeLabel ?? $this->attribute ?? 'value';
```
arogachev, line 250: Maybe make `value` customizable as well?

arogachev, line 250: What if `int` attribute names will be used? Will the result be like "0 is not valid."? Here the attribute name can be confused with actual value. Maybe turn it to countable items then? Like "1th item is not valid". Can be discussed in a separate issue and improved in a separate PR.

src/Rule/AbstractCompare.php
```diff
@@ -262,14 +262,14 @@ public function getIncorrectDataSetTypeMessage(): string
...
         return $this->message ?? match ($this->operator) {
-            '==', => 'Value must be equal to "{targetValueOrAttribute}".',
-            '===' => 'Value must be strictly equal to "{targetValueOrAttribute}".',
-            '!=' => 'Value must not be equal to "{targetValueOrAttribute}".',
-            '!==' => 'Value must not be strictly equal to "{targetValueOrAttribute}".',
-            '>' => 'Value must be greater than "{targetValueOrAttribute}".',
-            '>=' => 'Value must be greater than or equal to "{targetValueOrAttribute}".',
-            '<' => 'Value must be less than "{targetValueOrAttribute}".',
-            '<=' => 'Value must be less than or equal to "{targetValueOrAttribute}".',
+            '==', => '{Attribute} must be equal to "{targetValueOrAttribute}".',
```
arogachev, line 265: `targetValueOrAttribute` - support capitalization as well?

src/Rule/AtLeast.php
```diff
@@ -62,7 +62,7 @@ final class AtLeast implements DumpedRuleInterface, SkipOnErrorInterface, WhenIn
     public function __construct(
         private array $attributes,
         private int $min = 1,
-        private string $incorrectInputMessage = 'The value must be an array or an object.',
+        private string $incorrectInputMessage = '{Attribute} must be an array or an object.',
         private string $message = 'At least {min, number} {min, plural, one{attribute} other{attributes}} from this ' .
```
arogachev, line 66: `attributes` - support capitalization as well?

arogachev (changes requested):

src/ValidationContext.php
vjik: In what case may this be required?

vjik: The value cannot be `int`. For `int` keys attributes is `null`.

arogachev: Sorry, didn't get you. Could you demonstrate it with an example?

arogachev: It could be useful to customize text for "value" globally (for every error message). This can be completely another word or the same word with adjective added for example, something like "This value" and similar.

vjik: See here: https://github.com/dood-/validator/blob/enh-attribute-name-in-messages/src/Validator.php#L134-L142

samdark (approved):

samdark:
Thank you!

PR #665, opened 2024-02-24
Add method `Result::addErrorWithoutTranslation()`

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌
| Fixed issues  |


samdark:
What's the use-case for not translating errors?

vjik:
> What's the use-case for not translating errors?

1) In single language apps for custom error messages. Such messages not need to translate.

2) This PR prevent duplicate translations (after Each and Nested rules), because validator always returns translated messages.

samdark (approved):

samdark (approved):

PR #695, opened 2024-04-10
Add `Unique` rule

arogachev:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌

### DIFF
--- a/src/Rule/UniqueHandler.php
+++ b/src/Rule/UniqueHandler.php
@@ -1,0 +1,74 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Yiisoft\Validator\Rule;
+
+use DateTime;
+use DateTimeInterface;
+use Stringable;
+use Yiisoft\Validator\Exception\UnexpectedRuleException;
+use Yiisoft\Validator\Result;
+use Yiisoft\Validator\RuleHandlerInterface;
+use Yiisoft\Validator\ValidationContext;
+
+use function in_array;
+
+/**
+ * A handler for {@see Unique} rule. Validates uniqueness of each element of an iterable.
+ */
+final class UniqueHandler implements RuleHandlerInterface
+{
+    public function validate(mixed $value, object $rule, ValidationContext $context): Result
+    {
+        if (!$rule instanceof Unique) {
+            throw new UnexpectedRuleException(Unique::class, $rule);
+        }
+
+        if (!is_iterable($value)) {
+            return (new Result())->addError($rule->getIncorrectInputMessage(), [
+                'attribute' => $context->getTranslatedAttribute(),
+                'type' => get_debug_type($value),
+            ]);
+        }
+
+        $stackMap = ['simple' => [], 'datetime' => []];
+        foreach ($value as $item) {
+            if (!$this->isValueAllowedForItem($item)) {
+                return (new Result())->addError($rule->getIncorrectItemValueMessage(), [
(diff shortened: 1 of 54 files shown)


src/Rule/UniqueHandler.php
```diff
@@ -0,0 +1,74 @@
...
+                return (new Result())->addError($rule->getIncorrectItemValueMessage(), [
+                    'attribute' => $context->getTranslatedAttribute(),
+                    'type' => get_debug_type($item),
+                ]);
+            }
+
+            if ($item instanceof Stringable) {
+                $itemValue = (string) $item;
+                $stackMapKey = 'simple';
+            } elseif ($item instanceof DateTime) {
```
xepozz, line 47:
```suggestion
            } elseif ($item instanceof DateTimeInterface) {
```

src/Rule/UniqueHandler.php
```diff
@@ -0,0 +1,74 @@
...
+                $itemValue = $item->getTimestamp();
+                $stackMapKey = 'datetime';
+            } else {
+                $itemValue = $item;
+                $stackMapKey = 'simple';
+            }
+
+            if (in_array($itemValue, $stackMap[$stackMapKey], strict: true)) {
```
xepozz, line 55: Are you sure that's right way to check `$itemValue` not from the whole bucket?

src/Rule/UniqueHandler.php
```diff
@@ -0,0 +1,74 @@
...
+                return (new Result())->addError($rule->getMessage(), [
+                    'attribute' => $context->getTranslatedAttribute(),
+                ]);
+            }
+
+            $stackMap[$stackMapKey][] = $itemValue;
+        }
+
+        return new Result();
```
xepozz, line 64:
```suggestion
            $stackMap[$stackMapKey][] = $itemValue;
        }
        if (count(array_flip($stackMap)) !== count($stackMap)) {
            return (new Result())->addError($rule->getMessage(), [
                'attribute' => $context->getTranslatedAttribute(),
            ]);
        }

        return new Result();
```

Also duplicated items could be mentions in the error

src/Rule/UniqueHandler.php
arogachev: Fixed. :heavy_check_mark:

src/Rule/UniqueHandler.php
arogachev: 1. Not sure if `count` with `array_flip` are needed here, both logic and performance-wise. We only need to find the first occurrence and stop there. Also there are some requirements to handle some items in a specific way. I got rid of `in_array` in favor of a regular `foreach` loop.
2. It can be useful but at the same time it can cause unpredictable output in case of large values, etc.

vjik (commented):
May be use name "ArrayUnique"?

arogachev:
How about `UniqueIterable`?

vjik:
> How about `UniqueIterable`?

It's good for me.

src/Rule/UniqueHandler.php
xepozz: Could you bench positive test-cases with 2 loops and one + hashmap check?

arogachev: I replaced it with other approach with `count` and `array_unique`.

samdark (approved):

PR #747, opened 2024-09-07
Merge rules from PHP attributes with rules provided via `getRules()` method

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ✔️/❌

### DIFF
--- a/src/DataSet/ObjectDataSet.php
+++ b/src/DataSet/ObjectDataSet.php
@@ -5,15 +5,21 @@ declare(strict_types=1);
 namespace Yiisoft\Validator\DataSet;
 
 use ReflectionProperty;
+use Traversable;
 use Yiisoft\Validator\PropertyTranslatorInterface;
 use Yiisoft\Validator\PropertyTranslatorProviderInterface;
 use Yiisoft\Validator\DataSetInterface;
 use Yiisoft\Validator\DataWrapperInterface;
 use Yiisoft\Validator\Helper\ObjectParser;
 use Yiisoft\Validator\LabelsProviderInterface;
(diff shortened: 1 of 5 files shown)


samdark (approved):

src/DataSet/ObjectDataSet.php
```diff
@@ -218,17 +225,41 @@ public function getRules(): iterable
...
+             * @psalm-var list<RuleInterface> $value If `$key` is string, then `$value` is array of rules
+             * @see ObjectParser::getRules()
+             */
+
+            if (!isset($rules[$key])) {
+                $rules[$key] = $value;
+                continue;
+            }
+
+            $rules[$key] = is_iterable($rules[$key])
```
xepozz, line 257: If `$rules[$key]` isn't iterable, why does it mean that `$value` is?

Btw, this line must be benchmarked. I'd suggest to use `foreach` loop with `[]=` operator

vjik: > If `$rules[$key]` isn't iterable, why does it mean that `$value` is?

`$value` is always list of rules here. See https://github.com/yiisoft/validator/blob/master/src/Helper/ObjectParser.php#L197

> Btw, this line must be benchmarked. I'd suggest to use foreach loop with []= operator

https://github.com/vjik/php-unpacking-vs-foreach-bench

I think in this case better use unpacking.

arogachev (approved):

Tigrov (approved):

xepozz (approved):

PR #758, opened 2025-02-06
Minor refactoring with PHP 8.1 features usage

vjik:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Breaks BC?    | ❌


samdark (approved):

composer.json
```diff
@@ -83,7 +83,7 @@
         "allow-plugins": {
             "infection/extension-installer": true,
             "composer/package-versions-deprecated": true,
-            "yiisoft/config": true
+            "yiisoft/config": false
```
BoShurik, line 86: Why?

vjik: Composer plugin is not used here. It works on app level only. Here Yii Config needs for tests only.

PR #766, opened 2025-04-27
Add UUID validator

Sakhnovkrg:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ✔️
| Breaks BC?    | ❌


src/Rule/Uuid.php
```diff
@@ -0,0 +1,106 @@
...
+
+    public function getName(): string
+    {
+        return self::class;
+    }
+
+    /**
+     * Get a message used when the input it incorrect.
+     *
+     * @return string Error message
```
samdark, line 65:
```suggestion
     * @return string Error message.
```

src/Rule/Uuid.php
```diff
@@ -0,0 +1,106 @@
...
+        private string $incorrectInputMessage = '{Property} must be a string. {type} given.',
+        private string $message = 'The value of {property} is not a valid UUID.',
+        bool|callable|null $skipOnEmpty = null,
+        private bool $skipOnError = false,
+        private Closure|null $when = null,
+    ) {
+        $this->skipOnEmpty = $skipOnEmpty;
+    }
+
+    public function getName(): string
```
samdark, line 57: Need a test for it.

src/Rule/UuidHandler.php
```diff
@@ -0,0 +1,50 @@
...
+    public function validate(mixed $value, RuleInterface $rule, ValidationContext $context): Result
+    {
+        if (!$rule instanceof Uuid) {
+            throw new UnexpectedRuleException(Uuid::class, $rule);
+        }
+
+        $result = new Result();
+
+        if (!is_string($value)) {
+            return $result->addError($rule->getIncorrectInputMessage(), [
```
samdark, line 33: Need a test for it.

samdark (commented):
:+1: Please add a line for CHANGELOG and missing tests to have full coverage. Other than that, looks good to me.

src/Rule/UuidHandler.php
Sakhnovkrg: This case is already covered in dataValidationFailed() with cases like 'incorrect type, integer' and 'array'.

samdark (approved):

samdark:
Thank you!

PR #783, opened 2025-09-14
Treat uninitialized property as null

batyrmastyr:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

Treat uninitialized property as null to align with ArrayDataSet and inexisting properties handling.
```php
$parser = new ObjectParser(new class () {
    public int $a = 4;
    public int $uninitialized;
});

$this->assertNull($parser->getPropertyValue('not-existing')); // null 
$this->assertSame(4, $parser->getPropertyValue('a'));   // 4
$this->assertNull($parser->getPropertyValue('uninitialized')); // exception in main branch
```


src/Helper/ObjectParser.php
```diff
@@ -279,9 +279,16 @@ public function getLabels(): array
...
-            ? ($this->getReflectionProperties()[$property] ?? null)?->getValue($this->source)
-            : null;
+        if (!is_object($this->source)) {
+            return null;
+        }
+
+        $reflectionProperty = $this->getReflectionProperties()[$property] ?? null;
+        if ($reflectionProperty === null || !$reflectionProperty->isInitialized($this->source)) {
+            return null;
+        }
```
samdark, line 289: That would cost some performance. Any idea on how much?

batyrmastyr: Didn't test previously but from an adhoc [benchmark](https://sandbox.ws/en/shared/93) I see that
- `isInitialized` «cost» is about 60% of `getValue`
- first access is several times more pricey than next ones.

On a more close [benchmark imitating cache access](https://sandbox.ws/en/shared/38bbdfaf-2054-4b07-bfc7-f4dbb0d05709) you get +- 25% time slowdown.

samdark: How much is that in ms for ~100 properties? I think that's the reasonable maximum for a validator (except if it's used in batches).

batyrmastyr: @samdark same proportion `20,320 µsec` vs `29,920 µsec` [source](https://sandbox.ws/en/shared/639d2c22-d9e2-40be-80de-d1723cf4a310)

smirnov-e: Client can sent any part of data, can typos to property names and may be evil hacker.
The only alternatives I see are:
- use isset, but that will drop support of private and protected properties.
- suggest to completely drop property typing for incoming request parameters or use only constructor promoted properties.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## 2.5.1 under development
 
 - Bug #776: Take `PropertyTranslatorProviderInterface` into account during context validation (@vjik)
+- Enh #783: Default Intl locale in tests (@batyrmastyr)
```
vjik, line 6:
```suggestion
```

Tests changes don't required to adding here.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
...
+- Bug #783: Exception on uninitialized properties validation (@batyrmastyr)
```
vjik, line 7:
```suggestion
- Bug #783: Fix error on uninitialized properties validation (@batyrmastyr)
```

tests/Helper/ObjectParserTest.php
```diff
@@ -178,4 +178,18 @@ public function testDisabledCache(): void
         $this->assertNotSame($rules, $parser->getRules());
         $this->assertArrayNotHasKey($cacheKey, $cacheProperty->getValue());
     }
+
+    public function testUninitializedProperty(): void
```
vjik, line 182: Add also test for `(new Validator())->validate($objectWithUnitializedProperty)`

src/Helper/ObjectParser.php
```diff
@@ -279,9 +279,16 @@ public function getLabels(): array
      */
     public function getPropertyValue(string $property): mixed
```
vjik, line 280: What about `hasProperty()`? Should it returns true on non-initialized property or not?

vjik (changes requested):

src/Helper/ObjectParser.php
smirnov-e: I guess it should return true because property is defined, though this a bit contradicts with `->getData()` hiding such property.

batyrmastyr: I guess it should return true because property is defined
-  this a bit contradicts with `->getData()` hiding such property, but hydrating new object from `->getData` will not break on unexpected data.

src/Helper/ObjectParser.php
batyrmastyr: Yes, µsec

tests/Helper/ObjectParserTest.php
batyrmastyr: Added

src/Helper/ObjectParser.php
smirnov-e: And this behavior matches `get_object_vars` and `property_exists`.

batyrmastyr: And this behavior matches `get_object_vars` and `property_exists`.

vjik (approved):

PR #790, opened 2026-03-21
Fix #709: Use format for error message dates when unset

### LINKED ISSUE #709: In date rules use `format` property for format date in error message when don't set other format properties
For example:

```php
new Date(
    format: 'php:d.m.Y',
    min: '19.11.2013',
    max: '31.12.2099',
    skipOnEmpty: true,
),
```

In this case I expect that in error message date will be formatted as `php:d.m.Y`.

WarLikeLaux:
| Q             | A
| ------------- | ---
| Is bugfix?    | ❌
| New feature?  | ❌
| Tests added?  | ✔️
| Breaks BC?    | ✔️
| Fixed issues  | #709

## What does this PR do?

In date rules, use the `format` property for formatting dates in error messages when `messageFormat` and `messageDateType`/`messageTimeType` are not set.

Handler-level message types take precedence over the rule's `format`: an explicit `new TimeHandler(messageTimeType: IntlDateFormatter::FULL)` keeps formatting the limit even when the rule has a `format`. New `defaultMessageDateType`/`defaultMessageTimeType` parameters carry the previous `SHORT` default without overriding the rule, so omitting message types behaves as before.

### BC break

When `format` is set but `messageFormat`/`messageDateType`/`messageTimeType` are not, dates in error messages are now formatted using `format` instead of `IntlDateFormatter::SHORT`.

Before:
```php
new Date(format: 'php:d.m.Y', max: '31.12.2099');
// error: "Value must be no later than 12/31/99."
```

After:
```php
new Date(format: 'php:d.m.Y', max: '31.12.2099');
// error: "Value must be no later than 31.12.2099."
```

To keep the old behavior, set `messageFormat`, `messageDateType`, or `messageTimeType` explicitly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved date and time error message formatting to consistently use the defined format property when message-specific formatting properties are not set, ensuring clearer and more consistent validation error messages.

* **Tests**
  * Updated validation error message assertions to reflect new date/time formatting standards.

[Review Change Stack](https://app.coderabbit.ai/change-stack/yiisoft/validator/pull/790?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


samdark (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 2.5.2 under development
 
+- Chg #789: Use `format` property for formatting dates in error messages when message format properties are not set (@WarLikeLaux)
```
samdark:
```suggestion
- Chg #709: Use `format` property for formatting dates in error messages when message format properties are not set (@WarLikeLaux)
```

WarLikeLaux:
@coderabbitai full review

WarLikeLaux:
@coderabbitai full review

src/Rule/Date/BaseDateHandler.php
```diff
@@ -187,14 +187,20 @@ private function prepareValueWithIntlFormat(
...
             ?? $this->getDateTypeFromRule($rule);
-        $formatterTimeType = $this->getMessageTimeTypeFromRule($rule)
+        $formatterTimeType = $ruleMessageTimeType
             ?? $this->messageTimeType
             ?? $this->getTimeTypeFromRule($rule);
 
         $format = $rule->getMessageFormat() ?? $this->messageFormat;
+        if ($format === null && $ruleMessageDateType === null && $ruleMessageTimeType === null) {
+            $format = $rule->getFormat();
+        }
```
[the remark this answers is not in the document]
WarLikeLaux: Intentional. The fallback keys off the rule's own message properties, not the handler's - a `format` on the rule wins over handler-level `messageDateType`/`messageTimeType` defaults (that's the point of #709). To keep the handler's Intl style, set the message properties on the rule.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## 2.5.2 under development
 
+- Chg #709: Use `format` property for formatting dates in error messages when message format properties are not set (@WarLikeLaux)
```
vjik:
```suggestion
- Enh #709: In date rules use `format` property for formatting dates in error messages when message format properties are not set (@WarLikeLaux)
```

vjik (approved):

src/Rule/Date/BaseDateHandler.php
```diff
@@ -187,14 +187,20 @@ private function prepareValueWithIntlFormat(
...
+        $formatterDateType = $ruleMessageDateType
             ?? $this->messageDateType
             ?? $this->getDateTypeFromRule($rule);
-        $formatterTimeType = $this->getMessageTimeTypeFromRule($rule)
+        $formatterTimeType = $ruleMessageTimeType
             ?? $this->messageTimeType
             ?? $this->getTimeTypeFromRule($rule);
 
         $format = $rule->getMessageFormat() ?? $this->messageFormat;
+        if ($format === null && $ruleMessageDateType === null && $ruleMessageTimeType === null) {
```
samdark: This overrides explicit handler-level message formatting. For example, an app-level `new TimeHandler(messageTimeType: IntlDateFormatter::FULL)` is now ignored whenever the rule has `format`, because this fallback only checks the rule-level message type properties. That makes global message formatting configuration ineffective unless every rule also sets `messageTimeType` / `messageDateType` or `messageFormat`. We need to preserve handler-level precedence here; with the current constructor defaults, that may require distinguishing default handler message types from explicitly configured ones before falling back to `$rule->getFormat()`.

WarLikeLaux: @coderabbitai please verify fix

WarLikeLaux: @samdark got handler precedence working with the `SHORT` defaults preserved (BC), tests + mutation coverage green. But I'm relying on `func_num_args()` to distinguish a passed arg from the default, which is brittle - tied to arg positions. Not sure it's the right call. How would you approach distinguishing default vs explicit here, and do you want the explicit-`null`-follows-rule-type bit at all? Happy to redo it your way.

samdark: I don't think it's a good approach. Looks weird and is fragile. When named arguments are used, it breaks:

```php
new DateHandler(
    messageDateType: IntlDateFormatter::SHORT,
    tooLateMessage: 'Max: {limit}.',
);
```

samdark: I think the best approach is to change defaults to `null`. It is backwards compatibility break and can be in major release only but still that's better than super-hacky way with `func_num_args()`.

samdark: @vjik maybe you have a better idea?

vjik: We can try add `default*` parameters, that don't override rule parameters and use current parameters to override rule parameters.

WarLikeLaux: Reworked per @vjik - new `defaultMessageDateType`/`defaultMessageTimeType` keep the old `SHORT` default without overriding the rule, and `messageDateType`/`messageTimeType` are now the override over the rule's `format`. Dropped `func_num_args()`.

The BC checker flags the `messageDateType` default going from `SHORT` to `null`, but I think it's only the signature - `new *Handler()` formats exactly as before because `default*` carries the `SHORT`. So the report is red, the runtime behavior isn't.

WarLikeLaux: @coderabbitai please double-check the implementation and the BC impact

PR #803, opened 2026-06-04
Make file size messages human-readable

### LINKED ISSUE #802: Make File size validation messages human-readable
`Yiisoft\Validator\Rule\File` size validation placeholders currently expose raw byte counts.

Example:

```php
new File(
    maxSize: 50 * 1024 * 1024,
    tooBigMessage: "{file} is larger than {limit}.",
)
```

`{limit}` resolves to `52428800`, which is not convenient for user-facing upload validation messages.

Preferred behavior: use the same existing placeholders, but format size values in a human-readable way:

- `{limit}` -> `50 MB`
- `{exactly}` -> `50 MB`

This would let applications use Yii Validator defaults/placeholders without duplicating byte-to-readable-size formatting in application code.

samdark:
Resolves #802.

## Summary
- Format File size placeholders as human-readable values for {limit} and {exactly}.
- Update default File size messages to use the formatted placeholders directly.
- Add regression coverage and document the placeholder behavior.

## Tests
- vendor/bin/phpunit tests/Rule/FileTest.php --testdox
- vendor/bin/psalm --config=psalm.xml --output-format=console --no-cache --threads=1
- vendor/bin/php-cs-fixer fix --dry-run --diff --sequential --allow-unsupported-php-version=yes

Full vendor/bin/phpunit was also run; it fails only on existing EmailTest DNS-dependent cases for gmail.com and ñandu.cl domains in this environment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.6.1 under development
 
-- no changes in this release.
+- Bug #802: Use translatable human-readable file sizes in `File` validation messages and add numeric, unit, and raw byte placeholders (@samdark)
 
 ## 2.6.0 June 02, 2026
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.6.1 under development
 
-- Bug #802: Use translatable human-readable file sizes in `File` validation messages and add numeric, unit, and raw byte placeholders (@samdark)
+- Chg #802: Use translatable human-readable file sizes in `File` validation messages and add numeric, unit, and raw byte placeholders (@samdark)
 
 ## 2.6.0 June 02, 2026
 
--- a/src/Rule/FileHandler.php
+++ b/src/Rule/FileHandler.php
@@ -21,6 +21,7 @@ use function is_readable;
 use function is_string;
 use function mime_content_type;
 use function pathinfo;
+use function round;
 use function str_contains;
 use function str_ends_with;
 use function str_replace;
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## 2.6.1 under development
 
-- no changes in this release.
+- Bug #802: Use translatable human-readable file sizes in `File` validation messages and add numeric, unit, and raw byte placeholders (@samdark)
```
vjik:
```suggestion
- Chg #802: Use translatable human-readable file sizes in `File` validation messages and add numeric, unit, and raw byte placeholders (@samdark)
```

src/Rule/FileHandler.php
```diff
@@ -279,27 +293,160 @@ private function validateSize(array $file, File $rule, ValidationContext $contex
...
 
+    private function getNotExactSizeMessage(File $rule, string $unit): string
+    {
+        $message = $rule->getNotExactSizeMessage();
+        if (!in_array($message, self::DEFAULT_NOT_EXACT_SIZE_MESSAGES, true)) {
+            return $message;
+        }
+
+        return match ($unit) {
+            'byte' => 'The size of {property} must be exactly {exactlyValue, number} '
```
vjik, line 358: Add new variable `{prettyExactly}` and use it as default message in rule "The size of {property} must be exactly {prettyExactly}". Then insert the appropriate value in the handler.

This approach simplify code (not need constants with array of messages).

Even better use pretty value for `{exactly}` and create new `{exactlyHuman}`. So, it will be:

- `{exactly}` - pretty value
- `{exactlyHuman}` - expected exact size as a human-readable string.
- `{exactlyValue}`: expected exact size numeric value converted to a human-readable unit.
- `{exactlyUnit}`: expected exact size unit identifier. Possible values are `byte`, `KB`, `MB`, `GB`, `TB`, `PB`.
- `{exactlyBytes}`: expected exact size in bytes.

The same is true for new "limit" variables.

vjik (changes requested):

src/Rule/FileHandler.php
samdark: @vjik that won't work for pretty values since units are to be translated as well.

10 Mb -> 10 Мб
2 bytes -> 2 байта

PR #817, opened 2026-09-11
Fix #817: Fix `UniqueIterable` comparing the whole iterable instead of its items

CorvusSharp:
| Q             | A
| ------------- | ---
| Is bugfix?    | ✔️
| New feature?  | ❌
| Breaks BC?    | ❌

`UniqueIterableHandler` never compares the items of the iterable. Inside the loop the
stack is filled with `$value`, the whole iterable, instead of `$item`:

```php
if ($value instanceof Stringable) {
    $stack[] = (string) $value;
} elseif ($value instanceof DateTimeInterface) {
    $stack[] = $value->getTimestamp();
} else {
    $stack[] = $value;
}
```

So the stack holds N copies of one and the same iterable, and the uniqueness check fires
as soon as it holds two of them, regardless of the content.

The result is that the rule reports the opposite of the truth for most inputs:

| Value | Current | Expected |
| --- | --- | --- |
| `['a', 'a']` | valid | not unique |
| `['a', 'b', 'c']` | not unique | valid |
| `[1, 1]` | valid | not unique |
| `[1, 2, 3, 4]` | not unique | valid |

Any iterable with three or more items is reported as containing duplicates, and a real
duplicate in a pair passes.

There is a second, smaller defect in the same place: the check runs before the current
item is pushed, so the last item never takes part in it. Even with the right variable,
a duplicate formed by the final item would be missed.

The existing tests do not catch any of this because of how the data sets are shaped:
every case in `dataValidationPassed` has exactly two items, and every duplicate case in
`dataValidationFailed` has three or more. Both then pass for the wrong reason.

## The fix

- push `$item` into the stack instead of `$value`;
- run the uniqueness check once, after the loop, so every item takes part in it.

The order of the `Stringable` and `DateTimeInterface` branches is left as it was.
Swapping them would change how an object implementing both interfaces is compared, which
is unrelated to this bug.

### Order of errors

Moving the check out of the loop makes the order explicit and keeps per item errors
ahead of it: an item with a disallowed value or a type different from its predecessor is
reported wherever it occurs, and uniqueness is only reported when every item is valid.
Two data sets pin this down: `[1, 1, []]` reports the disallowed item and `[1, 1, 'a']`
reports the type mismatch, not the duplicate.

As a side effect `array_unique` now runs once instead of once per item.

## Tests

Thirteen data sets added:

- five in `dataValidationPassed`: unique lists of three and four items — strings,
  integers, floats, `Stringable`, `DateTimeInterface`;
- six in `dataValidationFailed`: two equal items — strings, integers, floats, booleans,
  `Stringable`, `DateTimeInterface`;
- two in `dataValidationFailed` for the order of errors described above.

The eleven covering the comparison itself fail on the current handler and pass with the
fix. The full suite stays green: 1945 tests, 4018 assertions. Psalm reports no issues.

## About the failing `rector` job

It fails on `master` too, so it is not related to this change. `rector.php` skips
`Rector\Php81\Rector\ClassMethod\NewInInitializerRector`, and that rule no longer exists
in Rector, so the run stops with "These rules from `$rectorConfig->skip()` do not exist".
Reproduced locally on a clean checkout of `master`; the last green run of that workflow
on `master` was in April. Removing the stale entry also uncovers two files the current
Rector wants to rewrite and one more skip that is never registered, so it looks like
separate housekeeping rather than something to fold in here. Happy to send that as its
own pull request if you want it.


CorvusSharp:
Good catch on the error order, that was a real inconsistency between the code and what I
wrote in the description.

Fixed by running the uniqueness check once after the loop instead of inside it. Per item
errors now always come first: an item with a disallowed value or a type different from
its predecessor is reported wherever it occurs, and uniqueness is reported only when
every item is valid. Two data sets pin it down, `[1, 1, []]` reports the disallowed item
and `[1, 1, 'a']` reports the type mismatch rather than the duplicate.

As a side effect `array_unique` now runs once rather than once per item.

vjik (approved):

samdark (approved):

samdark:
:+1:

